PR #7, opened 2023-03-21
Installing Django in the VM 

amitKaplansky:
Django installed in the VM.
Django is a software framework for writing web applications using the Python programming language.

Added two files: Pipefile and Pipefile.lock.
- Pipefile is used to store the high-level human friendly information.
- Pipefile.lock stores the detailed low-level information.

Added meetballs directory, the directory works as a Python module. 
In the directory you will find: 
- __init__.py
- asgi.py
- settings.py
- urls.py
- wsgi.py

Added a manage.py
This script is used to interact with Django to develop and test the application.


OhadSeg:
Seems fine Amit

Yuval-Vino:
Great job on providing such an informative description! However, on the other hand, the title could use more informative details.
Other than that everything looks good, great job

Yuval-Vino:
> hey, nice progress ! if i may.. for me to create the web app, i need the meat_balls_app to be created as well, can you add it please ? the line to create the app is :
> 
> pipenv run python manage.py startapp meet_balls_app

The included request doesn't seem necessary for achieving the desired result(from this PR). In general, it's recommended to keep PRs as small as possible to make it easier for reviewers to understand the changes and reduce the risk of errors or conflicts.

MeetBalls/__init__.py
yftacherzog, line 1: The `MeetBalls` directory is a python package, so it should be named according to the naming convention for python packages. Namely, it should be all lowercase (underscores can be used but are discouraged in this context, so `meetballs` or less preferably `meet_balls`).

yftacherzog (changes requested):
Other than the comment I left on the directory name, can you please change the commit message title so it will adhere to the structure mentioned on the contributing guide. Specifically, the line about using the imperative mood (e.g. `Add meetballs directory and ...` instead of `Adding ...`).

The fact that the PR is split into 2 commit might make your life a little harder for changing the first commit message. It's totally doable, but you can also squash this entire PR into one commit (I'd usually recommend it for various reasons) and make your life easier.

MeetBalls/__init__.py
amitKaplansky: I changed the commit as requested, should I change it to one commit only? way is it recommended?

yftacherzog: You don't have to change it. Personally, I very rarely include more than 1 commit on my PRs. It often means that the PR is too big (not in this case actually) or that the commits are not atomic, i.e. one cannot live without the other (also not the case here). 

All in all, it introduces headaches during code review and rarely provide much benefits.

manage.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+"""Django's command-line utility for administrative tasks."""
+import os
+import sys
+
+
+def main():
+    """Run administrative tasks."""
+    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'MeetBalls.settings')
```
yftacherzog, line 9: I suspect this needs to be `meetballs`. No?

meetballs/settings.py
```diff
@@ -0,0 +1,123 @@
...
+    'django.middleware.security.SecurityMiddleware',
+    'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.common.CommonMiddleware',
+    'django.middleware.csrf.CsrfViewMiddleware',
+    'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'django.contrib.messages.middleware.MessageMiddleware',
+    'django.middleware.clickjacking.XFrameOptionsMiddleware',
+]
+
+ROOT_URLCONF = 'MeetBalls.urls'
```
yftacherzog, line 52: `meetballs`?

Danielsio:
Seems like it can be merged now. good job.

meetballs/settings.py
amitKaplansky: As requested I generated the structure again.

Danielsio:
Structure looks good to me, i think it's ready to merge.

Yuval-Vino:
I see you addressed all the reviews, but this PR is still missing the relevant issue, as the contributing file demands.

PR #19, opened 2023-04-12
Add Court entity (Issue #18)

### LINKED ISSUE #18: Need to add Court entity

ZvikaNaorCohen:
- Add Court entity.
- Register Court in Admin file.
- Update models.py file to hold the relevant datamembers for Court entity.

closes #18


court/migrations/0001_initial.py
```diff
@@ -0,0 +1,20 @@
...
+
+    initial = True
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Court',
+            fields=[
```
Yuval-Vino, line 16: Why only the `id` field is mentioned here? Where are the rest of the fields?

ZvikaNaorCohen:
This PR is still being worked on, please don't review or comment it yet, as long as it's in draft mode. I'll let you guys know when it's ready. Thanks.

court/migrations/0001_initial.py
ZvikaNaorCohen: Fixed. Thank you for the reference.

ZvikaNaorCohen:
PR is ready to be reviewed. Thank you for your patience.

court/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Court(models.Model):
+    id = models.IntegerField(primary_key=True)
+    x = models.FloatField(default=0)
+    y = models.FloatField(default=0)
+    city = models.CharField(max_length=50, blank=True)
```
IdanZimi, line 8: From what I know, "blank = True" means the field is considered optional.
I personally think that city field is required. Please consider changing 'blank' attribute to false.

Danielsio (approved):

MaayanMashhadi (approved):

court/models.py
yftacherzog: I'd suggesting making as fewer optional fields as possible. It will make your life easier.

court/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Court(models.Model):
+    id = models.IntegerField(primary_key=True)
+    x = models.FloatField(default=0)
+    y = models.FloatField(default=0)
```
yftacherzog, line 7: What's the reason to have a default value here?

court/models.py
amitKaplansky: @yftacherzog do you think we should combine x and y together as a tuple?

court/models.py
```diff
@@ -0,0 +1,10 @@
...
+    city = models.CharField(max_length=50, blank=True)
+    neighborhood = models.CharField(max_length=50, blank=True)
+    max_players = models.PositiveIntegerField(default=0)
```
yftacherzog, line 10: Does this mean that if the court is initialized without specifying this field, then it has a max number of zero? Is this the expected behavior?

yftacherzog (changes requested):
Other than the comments, make sure the commit message follows the guidelines on the contributing guide (e.g. use the imperative mood).

Also, include a description to the PR on the first comment (other than just stating `closes ...`).

Typically, if you have only one commit on the PR (as you usually should), you can have the name of the PR the same as the commit title, and the PR description the same as the commit body + a line to link the PR with an issue.

court/models.py
yftacherzog: If I were you, I'd search for online references as I'm sure this is something others did in the past already. See an example of such discussion: https://stackoverflow.com/questions/30706799/which-model-field-to-use-in-django-to-store-longitude-and-latitude-values

ZvikaNaorCohen: Will be updated asap. Thanks.

court/models.py
ZvikaNaorCohen: When I wrote the default value, I was thinking about some habits from other courses. 
As for this PR, the mandatory data members are ID, X, Y, Max Players (city and neighborhood could have been blank).
So I was thinking that each mandatory field should have a default value, just because of that habit from other courses. I'm not saying that it's the best practice, but what is your opinion about default values in data members in general? Should we use them, or should we not?

court/models.py
ZvikaNaorCohen: It seems like I was not 100% clear of the attributes, and in second thought, the default attribute is unnecessary. Will be fixed. Thanks.

court/models.py
yftacherzog: I think default values should be used when it makes sense. Coordinates is something that is expected to be unique, so having a default value to it doesn't make sense to me.

In general, default values have the potential for introducing ambiguity, and ambiguity is something we're trying to avoid.

court/models.py
ZvikaNaorCohen: Fixed.

court/models.py
ZvikaNaorCohen: Thank you for your comment. I have fixed this part.

court/models.py
ZvikaNaorCohen: Have been updated. Thank you.

ZvikaNaorCohen:
I have updated the PR with the new comments.
Please leave your feedbacks regarding the changes.
I'll add the test parts hopefully tomorrow.

court/models.py
yftacherzog: Please make that change as part of the same original commit rather than creating a new commit.

yftacherzog:
> I have updated the PR with the new comments. Please leave your feedbacks regarding the changes. I'll add the test parts hopefully tomorrow.

Still need to fix the commit message (use the imperative mood)

ZvikaNaorCohen:
PR have been updated according to the last comments. 
Tests will be added later on today.
Please comment any feedback you have :-)

ZvikaNaorCohen:
Hello!
I've added tests to the Court entity, please let me know if you have any idea for a test that I haven't included.

tests/court_tests/test_court.py
```diff
@@ -0,0 +1,64 @@
...
+@pytest.fixture
+def court():
+    return Court(id=TEST_ID, x=TEST_X, y=TEST_Y, city=TEST_CITY, neighborhood=TEST_NEIGHBORHOOD,
+                 max_players=TEST_MAX_PLAYERS)
+
+
+@pytest.mark.django_db
+class TestCourtModel:
+
+    def test_court_model_save_method(self):
```
Yuval-Vino, line 22: this function should receive also `court` instead of creating one

tests/court_tests/test_court.py
```diff
@@ -0,0 +1,64 @@
...
+            max_players=TEST_MAX_PLAYERS
+        )
+        assert court.id == TEST_ID
+        assert court.x == TEST_X
+        assert court.y == TEST_Y
+        assert court.city == TEST_CITY
+        assert court.neighborhood == TEST_NEIGHBORHOOD
+        assert court.max_players == TEST_MAX_PLAYERS
+
+    def test_court_model_primary_key_field(self):
```
yftacherzog, line 38: Can you explain the rationale for this sort of test case? What are we testing here?

Yuval-Vino (changes requested):

tests/court_tests/test_court.py
```diff
@@ -0,0 +1,64 @@
...
+        court = Court(neighborhood='Hogwarts')
+        assert court.neighborhood == 'Hogwarts'
+
+    def test_court_model_max_players_field(self):
+        court = Court(max_players=5)
+        assert court.max_players == 5
+
+    def test_delete_court(self, court):
+        court.delete()
+        assert court not in Court.objects.all()
```
yftacherzog, line 64: You didn't check that it was not there in the first place, so how can you tell if it worked? I suspect it's not there as you didn't save it to the db.

tests/court_tests/test_court.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+from court.models import Court
+
+
+TEST_ID = 1
+TEST_X = 1.234567
+TEST_Y = 9.876543
+TEST_CITY = 'New York'
+TEST_NEIGHBORHOOD = 'Hogwarts'
+TEST_MAX_PLAYERS = 5
```
yftacherzog, line 10: I have some reservations about creating constants for test files on the global scope.

The reason being that those are not actually constants. They could be changed by one of the test cases and then affect all other test cases.

One approach is to define them as fixtures so each test case gets a fresh copy, but I'm not sure you actually need them defined in the global scope at this point.

tests/court_tests/test_court.py
```diff
@@ -0,0 +1,64 @@
...
+            city=TEST_CITY,
+            neighborhood=TEST_NEIGHBORHOOD,
+            max_players=TEST_MAX_PLAYERS
+        )
+        assert court.id == TEST_ID
+        assert court.x == TEST_X
+        assert court.y == TEST_Y
+        assert court.city == TEST_CITY
+        assert court.neighborhood == TEST_NEIGHBORHOOD
+        assert court.max_players == TEST_MAX_PLAYERS
```
yftacherzog, line 36: I think it will make more sense to fetch the instance from the db and then compare its fields to the object in the fixture.

pytest.ini
yftacherzog, line 1: add blank line at end of file

ZvikaNaorCohen:
I know that there should be only one commit in a single PR, however I encountered tiny bugs that I solved using GitHub instead of GIT. It caused 3 new commits that I can't squash. I've been checking guides on the internet but seems like nothing works. Has anyone encountered this and knows how to fix it? Or perhaps should I keep it like this?

yftacherzog:
> I know that there should be only one commit in a single PR, however I encountered tiny bugs that I solved using GitHub instead of GIT. It caused 3 new commits that I can't squash. I've been checking guides on the internet but seems like nothing works. Has anyone encountered this and knows how to fix it? Or perhaps should I keep it like this?

Don't leave it like this. 

What I'd suggest is that you:
1. Fetch upstream
2. Start an interactive rebase over upstream/main
3. Within the rebase, remove that merge commit. 
4. Either inside the same rebase or afterwards, squash the other 2 commits into the first one. 

If you can't figure it out, ping me tomorrow and we'll figure it out together.

ZvikaNaorCohen:
Yftach thank you for your help. I've managed to squash the commits, but suddenly Daniel's commit appears in mine. Moreover, the conflicting file still appears (settings.py) even tho I upstreamed, fetched, and updated the files as I believe I'm supposed to.
But the conflict won't dissapear.

yftacherzog:
> Yftach thank you for your help. I've managed to squash the commits, but the conflicting file still appears (settings.py) even tho I upstreamed, fetched, and updated the files as I believe I'm supposed to.

You're almost there.

1. checkout your PR branch (I assume it's in sync with what's here)
2. `git fetch upstream`
3. `git rebase upstream/main` -> it should now complain on a conflict. on the file you mentioned.
4. edit the conflicting file. You should see something like this:
```
<<<<<<< HEAD
=======
    'court.apps.CourtConfig',
>>>>>>> 83a7889 (Create Court entity)
```
It means that upstream/main has nothing on this line and you have your additions.
5. In this case, you just want to keep your line, so just delete the markers (the `<<<`, `====`, `>>>`).
6. Save the file.
7. `git add meetballs/settings.py`
8. `git rebase --continue`

It should now tell you that the rebase is done. Run git log to confirm that you can see the tip of upstream/main on your history. It means you're properly rebased.

use `git show` to see the content of your single commit and confirm this is what you expected the commit to include.

Force push.

ZvikaNaorCohen:
That's exactly what I've done earlier. Just with a small change - after continuing the rebase, I've committed the changes and then force pushed. 
And now I've tried according to your guide, and it solved the conflict.
Do you know why my redundant commit caused that conflict?

I think that my PR is finally ready to be reviewed. I've tested all the cases I could think about. If there is any question or feedback please let me know.

yftacherzog:
> That's exactly what I've done earlier. Just with a small change - after continuing the rebase, I've committed the changes and then force pushed. And now I've tried according to your guide, and it solved the conflict. Do you know why my redundant commit caused that conflict?
> 
> I think that my PR is finally ready to be reviewed. I've tested all the cases I could think about. If there is any question or feedback please let me know.

You don't commit during a rebase process unless you want to add a new commit (although it's a very common mistake as we're very used to commit after we add).

court/tests.py
```diff
@@ -0,0 +1,107 @@
...
+        assert court_in_db == court
+
+    def test_create_court_with_existing_id(self, court):
+        with pytest.raises(IntegrityError):
+            Court.objects.create(courtID=1, x=Decimal('11'), y=Decimal('22'),
+                                 city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
+
+    def test_create_valid_court(self):
+        Court.objects.create(courtID=1, x=Decimal('11'), y=Decimal('22'),
+                             city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
```
yftacherzog, line 27: use `full_clean` here to have the instance validated. See the player tests for examples.

court/tests.py
yftacherzog, line 1: What I'm not certain about is why on @Danielsio's PRs we had to use `full_clean` to have the error raised. Can you guys spot the differences? @Yuval-Vino do you know?

tests/court_tests/test_court.py
yftacherzog, line 1: Do we have the same tests on 2 different files? Let's make it one file and make sure we follow the same convention for the location as used for player.

ZvikaNaorCohen: Before Daniel's merge, I was thinking that we should have tests folder and each entity would have it's own folder. After his merge, I saw that he wrote the tests in the entity's folder, inside tests.py file.
The thing is, between all the mess that happened in this PR, I'm sure I've deleted tests/court_tests/test_court.py. I want the tests to be on court/tests.py. 
I'll update the PR and double check that tests/court_tests/ is being deleted.

court/tests.py
Yuval-Vino: A few things I notice 
`full_clean `raises `ValidationError` this type of error is used in Player PR, here we use `IntegrityError` (maybe caused by other reasons) 
Also, player fields (user, date, BallGame) have their own constraints (user must be FK, the date must be date type and BallGame have its choices) by Django
while in the Court entity there are no constraints by django on the fields (city, neighborhood,x,y...), possible that Django constraints on player fields raise the error

yftacherzog: @ZvikaNaorCohen let's have a test case for a court with city named `"a" * 60`, and see if we need to call clean_all for an error to be raised. if that is the case, let's add some test cases for the the constraints.

tests/court_tests/test_court.py
ZvikaNaorCohen: I agree. Again, I was sure that this folder doesn't exist. I'm uploading now a new PR without this folder. Thanks for the reference.

court/tests.py
ZvikaNaorCohen: If I've understood correctly, I added this test: 
image
and it passes without any errors. That's the expected behaviour, right?

court/tests.py
```diff
@@ -0,0 +1,107 @@
...
+                x=Decimal('1'),
+                city='TEST_CITY',
+                neighborhood='TEST_NEIGHBORHOOD',
+                max_players=10
+            )
+
+    def test_court_without_city(self, court):
+        with pytest.raises(IntegrityError):
+            Court.objects.create(
+                courtID=1,
```
Yuval-Vino, line 82: I think the error here is raised because your trying to create a court with existing courtID, not because of a blank city field

court/tests.py
```diff
@@ -0,0 +1,107 @@
...
+                x=Decimal('1'),
+                y=Decimal('1'),
+                neighborhood='TEST_NEIGHBORHOOD',
+                max_players=10
+            )
+
+    def test_court_without_neighborhood(self, court):
+        with pytest.raises(IntegrityError):
+            Court.objects.create(
+                courtID=1,
```
Yuval-Vino, line 92: also here

court/tests.py
```diff
@@ -0,0 +1,107 @@
...
+                x=Decimal('1'),
+                y=Decimal('1'),
+                city='TEST_CITY',
+                max_players=10
+            )
+
+    def test_court_without_max_players(self, court):
+        with pytest.raises(IntegrityError):
+            Court.objects.create(
+                courtID=1,
```
Yuval-Vino, line 102: also here

Yuval-Vino (changes requested):

court/tests.py
ZvikaNaorCohen: After changing the courtID to 800 (which doesnt exist if I'm not mistkaen) the tests remains the same - Integrity Error is raised.

court/tests.py
yftacherzog: > If I've understood correctly, I added this test: image (63 times A) 

`city="A" * 63` will make the intention clearer and the screen cleaner. Another approach will be to have something like `city="A city with a very long name which is longer than 50 chars".

When I tried to add .clean_all() it says Court object has no attribute named clean_all. When I tried to add .full_clean() it raised ValidationError.
> 
> So you suggest I'll add more test cases including the .full_clean cmd?

Yes. Just make sure to include only one violation on each test case so we can tell which is which (have a separate test function for each field).

court/tests.py
yftacherzog: The best approach will be to have a positive test case that works, and then negative test cases, each only changes one field comparing to the one that works. This way you know for sure why you failed.

Yuval-Vino: Another thing, I don't think you want  users to define their court_ID (how would they know if it's valid?), it should be assigned automatically with a legal value (which will happen if you use Django defaults PK)

tests/court_tests/test_court.py
ZvikaNaorCohen: Deleted this testcase.

tests/court_tests/test_court.py
ZvikaNaorCohen: Deleted. Thank you for your comment.

yftacherzog:
@ZvikaNaorCohen there's not a lot of work left here, and there are multiple levels of PRs that are dependent on this one. I'd suggest rebasing and addressing the open comment as soon as possible so it could be merged in time for the other PR to finalize their work.

court/tests.py
ZvikaNaorCohen: Will be added in the next PR. Thanks.

ZvikaNaorCohen:
> @ZvikaNaorCohen there's not a lot of work left here, and there are multiple levels of PRs that are dependent on this one. I'd suggest rebasing and addressing the open comments as soon as possible so it could be merged in time for the other PR to finalize their work.

Thank you for your comment. I've rebased and been working on the validation/integrity errors, I believe that all is ok now. Please lmk if there are any missing tests in your opinion.

court/tests.py
yftacherzog: Next PR?

yftacherzog:
> > @ZvikaNaorCohen there's not a lot of work left here, and there are multiple levels of PRs that are dependent on this one. I'd suggest rebasing and addressing the open comments as soon as possible so it could be merged in time for the other PR to finalize their work.
> 
> Thank you for your comment. I've rebased and been working on the validation/integrity errors, I believe that all is ok now. Please lmk if there are any missing tests in your opinion.

Did you push those changes? I see 2 force pushes but with changes which I think are unrelated.

court/tests.py
```diff
@@ -0,0 +1,112 @@
...
+from django.db.utils import IntegrityError
+from django.core.exceptions import ValidationError
+
+
+@pytest.mark.django_db
+class TestCourtModel:
+
+    @pytest.fixture
+    def court(self):
+        court = Court.objects.create(courtID=1, x=Decimal('11'), y=Decimal('22'),
```
yftacherzog, line 13: Following @Yuval-Vino's comment, do you need the court ID explicitly defined here? (also applies to `test_create_court_with_existing_id` and `test_create_valid_court`)

court/tests.py
```diff
@@ -0,0 +1,112 @@
...
+                                     city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
+        return court
+
+    def test_validate_instance(self, court):
```
yftacherzog, line 17:
```suggestion
    def test_get_instance(self, court):
```
You have another test case in which you validate the instance.

court/tests.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def test_create_court_with_negative_max_players(self):
+        with pytest.raises(IntegrityError):
+            Court.objects.create(
+                x=Decimal('11'),
+                y=Decimal('22'),
+                city='TEST_CITY',
+                neighborhood='TEST_NEIGHBORHOOD',
+                max_players=-100
+            ).full_clean()
```
yftacherzog, line 38: Is this exception raised by the call to full_clean (I think  the summary of the conversation from yesterday was that it is not)? If not, then the test should pass if you remove the call to full_clean.

court/tests.py
```diff
@@ -0,0 +1,107 @@
...
+            Court.objects.create(
+                courtID=1,
+                x=Decimal('11'),
+                y=Decimal('22'),
+                city='TEST_CITY',
+                neighborhood='TEST_NEIGHBORHOOD',
+                max_players=-100
+            )
+
+    def test_update_court(self, court):
```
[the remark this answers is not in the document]
yftacherzog: I'll resolve this one once you add a link to the new issue.

yftacherzog (changes requested):
Almost there!

court/tests.py
Yuval-Vino: Also, what is the courtID value if you don't explicitly define it? are we sure that it will be auto-created with a unique value? 
Django default has this hyperparameter which is missing here `auto_created=True`. 
It's possible that the default value of auto_created is true but I think we need to check it

ZvikaNaorCohen: I've deleted the courtID from other tests, and these went pretty good. I forgot about these functions that you've commented on. Unfortunately I have some errors when deleting the courtID, I'll make it work today.
@Yuval-Vino I've seen on a comment on StackOverflow mentioning that it's recommended to not use this hyperparameter, but the comment was from 2015 so I guess it's irrelevant. 
I'll fix these functions today. Thank you guys for your comments.

court/tests.py
ZvikaNaorCohen: I meant that it would be updated next time I'll amend commit updates. 
I'm confused when we want to add `.full_clean()` and when not. I've been googling about it a bit but just got more confused. Could you please explain it to me?

court/tests.py
ZvikaNaorCohen: You're right, I've totally forgot to delete the `.full_clean()` cmd. The `IntegrityError` occurs without it.

ZvikaNaorCohen:
I've changed my models file so that if you don't specify courtID, you'll be assigened one.
There are tests to make sure that you won't be able to create a court with existing ID, even if you add one by your own.
PR is ready to be reviewed.

court/models.py
amitKaplansky, line 1: In max_players field you assigned it be a positive Integer field which means this field could also be 0, shouldn't it be at least 1?

ZvikaNaorCohen: That's a nice point. 
I was thinking that should count from 0, in case one court goes under temporary maintenance (for example: A School's court, when we know that the court would be inaccessible for a short period of time).

amitKaplansky: Makes sense, I agree

Danielsio (approved):
Well Done Naor. I Agree with Amit, this PR can be merged.

Yuval-Vino (commented):
Great job Zvika!!! 
only one small note, i think you need to remove all comments from test code.
A test should tell the story without additional comments, you can create functions and variables with meaningful names that could help to explain what it does (self explanatory).

Yuval-Vino (approved):

court/models.py
```diff
@@ -0,0 +1,17 @@
...
+    y = models.DecimalField(max_digits=9, decimal_places=6)
+    city = models.CharField(max_length=50)
+    neighborhood = models.CharField(max_length=50)
+    max_players = models.PositiveIntegerField()
+
+    def save(self, *args, **kwargs):
+        if not self.courtID:
+            # generate a new unique courtID
+            max_court_id = Court.objects.aggregate(models.Max('courtID'))['courtID__max'] or 0
+            self.courtID = max_court_id + 1
```
yftacherzog, line 16: I don't think this is actually safe in terms of race conditions (what happens if two courts are created at the same time?), but I'm going to let it slide :)

ZvikaNaorCohen:
> Great job Zvika!!! only one small note, i think you need to remove all comments from test code. A test should tell the story without additional comments, you can create functions and variables with meaningful names that could help to explain what it does (self explanatory).

Thank you for your comment. 
I Updated the tests file and removed the comments.

court/tests.py
```diff
@@ -0,0 +1,129 @@
...
+                             city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30).full_clean()
+
+    def test_create_three_courts_without_id(self):
+        court_id1 = Court.objects.create(x=Decimal('11'), y=Decimal('22'),
+                                         city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
+        court_id2 = Court.objects.create(x=Decimal('22'), y=Decimal('33'),
+                                         city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=40)
+        court_id3 = Court.objects.create(x=Decimal('33'), y=Decimal('44'),
+                                         city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=50)
+        assert court_id1.courtID == 1
```
yftacherzog, line 37: There's a built-in frugality in this sort of test as you assume unnecessary things about the environment. In this case, you assume the first one will get ID of 1, but suppose that something changes in the setup and now you start the test with a court ID already in the system, so your test gets broken.

I think it would have been better to create the first one, grab the ID and then assert that the next one is its plus 1.

court/tests.py
```diff
@@ -0,0 +1,129 @@
...
+        with pytest.raises(IntegrityError):
+            Court.objects.create(courtID=1, x=Decimal('11'), y=Decimal('22'),
+                                 city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
+
+    def test_create_valid_court(self):
+        Court.objects.create(x=Decimal('11'), y=Decimal('22'),
+                             city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30).full_clean()
+
+    def test_create_three_courts_without_id(self):
+        court_id1 = Court.objects.create(x=Decimal('11'), y=Decimal('22'),
```
yftacherzog, line 31: Should be named court rather than court_id

court/tests.py
ZvikaNaorCohen: @yftacherzog I wasn't thinking of that option. You're right.
I'll open a new issue and fix it ASAP.

PR #21, opened 2023-04-12
Add player class entity

### LINKED ISSUE #20: Need to add Player entity

Danielsio:
I Created the player app with the starting files
Added the Player model to the player.models & the playerConfig to the settings
Declared 2 types of Enums in player.enums.py

closes #20


Danielsio:
Please have a look at the player.migrations.0002_test.py and give me notes if I need to change anything.

ZvikaNaorCohen:
The whole PR looks very good. One thing that I can't review yet is the tests file. After Sunday's lecture about Unit Tests I can re-read this file and comment.

Danielsio:
about the player model since it has a oneToOne field of the user, does it needs another field of player_id ?
or is it unnecessary?

player/models.py
```diff
@@ -0,0 +1,15 @@
+from django.db import models
+from django.conf import settings
+from player.enums import BallGame
+
+
+class Player(models.Model):
+    user = models.OneToOneField(settings.AUTH_USER_MODEL,
+                                on_delete=models.CASCADE)
```
Danielsio, line 8: should it have a player_id as well ?

player/models.py
```diff
@@ -0,0 +1,15 @@
...
+    birth_date = models.DateField(null=True, blank=True)
+    favorite_ball_game = models.CharField(
+        max_length=100,
+        choices=[(tag, tag.value) for tag in BallGame],
```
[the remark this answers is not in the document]
Danielsio: done.

player/models.py
Danielsio: so it's mandatory to add an id field as well? 
but i thought that using the oneToOne User field will make it so the user fields will be added to the model.. guess i will conduct some more reserach

player/enums.py
```diff
@@ -0,0 +1,30 @@
...
+    BASEBALL = 'Baseball'
+    TENNIS = 'Tennis'
+    RUGBY = 'Rugby'
+    GOLF = 'Golf'
+    CRICKET = 'Cricket'
+    HANDBALL = 'Handball'
+
+    @classmethod
+    def BALL_GAME_CHOICES(cls):
+        return [(tag, tag.value) for tag in cls]
```
Yuval-Vino, line 17: > Think about adding a function to the BallGame class called `BALL_GAME_CHOICES` that will return this array instead of using a list comprehension here. Alternatively, you could explore the use of `models.TextChoices` from the `django.db` module.
> 
> Edit: I just noticed that you define each value as a tuple of `tag `and `tag.value`. Is this the format that we want to store in the database, or should we only store the value and retrieve the tag in the code later when needed?

Moving the edit here (edited at the same time you updated the code)

player/models.py
Yuval-Vino: I see that you found a solution by specifying the user as pk

MaayanMashhadi:
I just see your changes and after comparing them to the ERD, It's look good for me.
I don't see any issues.

player/models.py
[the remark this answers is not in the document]
Danielsio: i get this error when i tried to change it to only store the value itself:

'choices' must be an iterable containing (actual value, human readable name) tuples.

player/enums.py
Danielsio: I got this error when i tried to change it to only store the value itself:

'choices' must be an iterable containing (actual value, human readable name) tuples.

MaayanMashhadi: Hi Daniel,
Try to switch the choices to tuple (with () and not []).
I search on "stack overflow" and this was one of the solution. Maybe it should fix the problem

Danielsio: Unfortunately this solutions didnt work for me.

yftacherzog:
> I just see your changes and after comparing them to the ERD, It's look good for me. I don't see any issues.

> The whole PR looks very good. One thing that I can't review yet is the tests file. After Sunday's lecture about Unit Tests I can re-read this file and comment.

@ZvikaNaorCohen @MaayanMashhadi 
This is still a lgtm comment to me. I can't possibly think that 10 files are changing (that's a big PR) and that's the only thing to say.

player/models.py
```diff
@@ -0,0 +1,18 @@
...
+from django.conf import settings
+from meet_balls_app.enums import BallGame
+
+
+class Player(models.Model):
+    user = models.OneToOneField(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        primary_key=True
+        )
```
yftacherzog, line 11: We're going to need unit tests to have this merged, as it includes integration with the built in user model and we need to see that the built in user is created/deleted/updated as expected.

player/models.py
yftacherzog, line 1: django documentation (see link below) is stating that you typically should not use `null=True`. Specifically, I think it will make you life easier for now if you set fields as mandatory unless you specifically need them as non-mandatory. 

https://docs.djangoproject.com/en/4.2/ref/models/fields/#null

meet_balls_app/enums.py
```diff
@@ -0,0 +1,30 @@
...
+    GOLF = 'Golf'
+    CRICKET = 'Cricket'
+    HANDBALL = 'Handball'
+
+    @classmethod
+    def BALL_GAME_CHOICES(cls):
+        return [(tag.value, tag.name) for tag in cls]
+
+
+class Rating(IntEnum):
```
yftacherzog, line 20: How is this one being used?

amitKaplansky: As Yuval suggested we added this instead in the models.py file:

class BallGame(models.TextChoices):
    Soccer = 'Soccer', 'Soccer'
    Basketball = 'Basketball', 'Basketball'
    Volleyball = 'Volleyball', 'Volleyball'
    Baseball = 'Baseball', 'Baseball'
    Tennis = 'Tennis', 'Tennis'
    Rugby = 'Rugby', 'Rugby'
    Golf = 'Golf', 'Golf'
    Cricket = 'Cricket', 'Cricket'
    Handball = 'Handball', 'Handball'

yftacherzog: @amitKaplansky Is that about the rating or about the game choices?

amitKaplansky: game choices, @MaayanMashhadi and I used it in the GameEvent entity

yftacherzog: My comment was regarding `class Rating`.

yftacherzog: @amitKaplansky Where is that entity that you mention?

amitKaplansky: still working on this entity,  so I didn't add the PR

amitKaplansky: I think line 20 can be deleted for now @Danielsio

Danielsio: > How is this one being used?

It's not being used currently, Removed.

amitCohen2:
Looks fine, waiting for your tests.
Good luck:)

meet_balls_app/enums.py
```diff
@@ -0,0 +1,17 @@
...
+    VOLLEYBALL = 'Volleyball'
+    BASEBALL = 'Baseball'
+    TENNIS = 'Tennis'
+    RUGBY = 'Rugby'
+    GOLF = 'Golf'
+    CRICKET = 'Cricket'
+    HANDBALL = 'Handball'
+
+    @classmethod
+    def BALL_GAME_CHOICES(cls):
```
amitCohen2, line 16: function name should be in lower case like the convention.
Also add new line at the end of this file.

meet_balls_app/enums.py
Danielsio: > As Yuval suggested we added this instead in the models.py file:
> 
> class BallGame(models.TextChoices): Soccer = 'Soccer', 'Soccer' Basketball = 'Basketball', 'Basketball' Volleyball = 'Volleyball', 'Volleyball' Baseball = 'Baseball', 'Baseball' Tennis = 'Tennis', 'Tennis' Rugby = 'Rugby', 'Rugby' Golf = 'Golf', 'Golf' Cricket = 'Cricket', 'Cricket' Handball = 'Handball', 'Handball'

Thanks you for the comment, that works.

meet_balls_app/enums.py
Danielsio: Thank you for the comment, it's fixed now.

Yuval-Vino: I am a bit confused, according to Amit you have implemented BallGame with TextChoices
In this implementation, you receive the choices function directly from Django, so there is no need to implement it

But here I see that it is implemented as an Enum
Can you clarify the final implementation for me?

Danielsio: > I am a bit confused, according to Amit you have implemented BallGame with TextChoices In this implementation, you receive the choices function directly from Django, so there is no need to implement it
> 
> But here I see that it is implemented as an Enum Can you clarify the final implementation for me?

I thought Amit refered to the comment where you hinted that we want to store only the tag.value in the database, so to make it work i figured that for each tag.value we display "Soccer", "Soccer" for example. 

I am not really sure myself what's the best way to declare this enum.

player/models.py
Danielsio: Fixed.

yftacherzog: In light of another conversation from this morning, I'd like to clarify that when saying making things mandatory, I meant not allowing them to be blank/null and not providing default values. i.e. making it mandatory for the user to specify a value explicitly.

Danielsio: The reason i left that as it is, was beacuse i got some weird error about no default value for ball game while trying to run makemigration.
now i fixed it and the error didnt show up so i am guessing something else was responsible for this error.. so anyway it's fixed now.

Danielsio:
@yftacherzog  @Yuval-Vino 
I Added some basic test at player/test.py. 

Please feel free to tell me if something else is needed.

player/tests.py
```diff
@@ -0,0 +1,38 @@
...
+def test_update_player(player):
+    player.favorite_ball_game = BallGame.Basketball
+    player.save()
+    updated_player = Player.objects.get(pk=player.pk)
+    assert updated_player.favorite_ball_game == BallGame.Basketball
+
+
+def test_delete_player(player):
+    player.delete()
+    assert not Player.objects.filter(pk=player.pk).exists()
```
yftacherzog, line 38: What happens if you run this assertion before calling delete? I suspect that this player is not saved (not sure if it's the same case for the user as well). You can maybe add another fixture `saved_player` in which you'd save the player, and then have the tests use that fixture.

player/tests.py
```diff
@@ -0,0 +1,38 @@
...
+
+def test_update_player(player):
+    player.favorite_ball_game = BallGame.Basketball
+    player.save()
+    updated_player = Player.objects.get(pk=player.pk)
+    assert updated_player.favorite_ball_game == BallGame.Basketball
+
+
+def test_delete_player(player):
+    player.delete()
```
yftacherzog, line 37: Does the user should be deleted as well? You have `on_delete=django.db.models.deletion.CASCADE` on the user field on the player definition. Does this mean that we should expect the user to be deleted? If so, need to test that.

player/tests.py
```diff
@@ -0,0 +1,38 @@
...
+@pytest.fixture
+def player(user):
+    return Player.objects.create(
+        user=user,
+        birth_date='1990-01-01',
+        favorite_ball_game=BallGame.Soccer
+    )
+
+
+def test_create_player(player, user):
```
yftacherzog, line 23: I'd expect to have a function that will encapsulate the creation of the user when creating the player. If this is something that you'd prefer doing later due to time limitations, then you can create an issue for it and do it after you're ready for the demo.

player/tests.py
yftacherzog: It's actually the other way around. The player should be deleted if the underlying user is being deleted. We should test that as well.

player/tests.py
yftacherzog: In any case, you have to be convinced that the player is actually saved.

player/tests.py
```diff
@@ -0,0 +1,46 @@
...
+        assert player.user == user
+        assert player.birth_date == '1990-01-01'
+        assert player.favorite_ball_game == BallGame.Soccer
+
+    def test_update_player(self, saved_player):
+        player = saved_player
+        player.favorite_ball_game = BallGame.Basketball
+        player.save()
+        updated_player = Player.objects.get(pk=player.pk)
+        assert updated_player.favorite_ball_game == BallGame.Basketball
```
yftacherzog, line 36: You need this sort of assertion when creating a player as well.

yftacherzog: more precisely. You need to verify that you can pull the player from the db and that it's the same player as you were expecting to find.

player/tests.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @pytest.fixture
+    def user(self):
+        return get_user_model().objects.create_user(
+            username='testuser',
+            password='testpass'
+        )
+
+    @pytest.fixture
+    def saved_player(self, user):
```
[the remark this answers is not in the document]
Danielsio: calling Player.objects.create() doesn't save the instance to the db ?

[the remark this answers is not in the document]
yftacherzog: I suggest adding a test for fetching the created instance from the db and making sure it's the same as the one in the fixture. If it's saved automatically, it will pass. Otherwise, save it to make it pass.

player/tests.py
Danielsio: added this assertion: 
        assert updated_player.pk == saved_player.pk
how's that ?

> more precisely, you need to verify that you can pull the player from the db and that it's the same player as you were expecting to find.

I am pulling the saved player by it's id so how could i get other player ?

player/tests.py
```diff
@@ -0,0 +1,47 @@
...
+        assert player.user == user
+        assert player.birth_date == '1990-01-01'
+        assert player.favorite_ball_game == BallGame.Soccer
+
+    def test_update_player(self, saved_player):
+        player = saved_player
+        player.favorite_ball_game = BallGame.Basketball
+        player.save()
+        updated_player = Player.objects.get(pk=player.pk)
+        assert updated_player.pk == saved_player.pk
```
yftacherzog, line 36: That was not my intention. you're fetching it by its pk, so you don't need to assert the pk. You don't need this line here.

player/tests.py
yftacherzog: This comment was not about this test, but about the test in which you create a player. There, you need to verify that you're able to fetch the instance.

player/tests.py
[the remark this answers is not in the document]
Danielsio: done

player/tests.py
Danielsio: > Does the user should be deleted as well? You have `on_delete=django.db.models.deletion.CASCADE` on the user field on the player definition. Does this mean that we should expect the user to be deleted? If so, need to test that.

I am kinda getting lost, i am pretty sure when user is deleted the player should be deleted as well, beacsue of the oneToOne field connection they have.. but this should be also true when player deleted the same user sould be deleted as well.

player/tests.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def test_delete_player(self, saved_player):
+        player = saved_player
+        player.delete()
+        assert not Player.objects.filter(pk=player.pk).exists()
+
+    def test_delete_user_deletes_player(self, saved_player):
+        user = saved_player.user
+        user.delete()
+        assert not Player.objects.filter(pk=saved_player.pk).exists()
```
yftacherzog, line 50: I'd make this assertion as closest as possible for being the exact opposite of the one verifying that the player was created. This way, you're asserting in the same way that it existed and that it doesn't exist once you remove it.

I think you should get an exception if you try to fetch it by its pk, so just the fetching inside the relevant context manager. Something like:

```suggestion
        with pytest.raises(TheTypeOfTheExpectedException):
                Player.objects.get(pk=saved_player.pk)
```

yftacherzog: Making the two checks different always raise the concern whether you actually changed something. i.e. you need to be convinced that this line did not return the same value before you called delete.

player/tests.py
yftacherzog: As far as I understand it, the cascade stuff means that deleting the user should also delete the player. Add a test to verify that.

player/tests.py
```diff
@@ -0,0 +1,93 @@
...
+        player = saved_player
+        player.favorite_ball_game = BallGame.Basketball
+        player.save()
+        updated_player = Player.objects.get(pk=player.pk)
+        assert updated_player.favorite_ball_game == BallGame.Basketball
+
+    def test_delete_player(self, saved_player):
+        player = saved_player
+        player.delete()
+        assert not Player.objects.filter(pk=player.pk).exists()
```
yftacherzog, line 46: I suggest this approach as you're using it to verify that the player does exist when it should.
```suggestion
        with pytest.raises(Player.DoesNotExist):
            Player.objects.get(pk=saved_player.pk)
```

player/tests.py
yftacherzog: Can you create an issue for that? I think this sort of method will be needed later on, but we can implement it when needed. I don't want it to hold this PR at this point, but I don't want it to be forgotten. You can resolve the conversation once you open the issue.

player/tests.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def test_get_player(self, saved_player):
+        player = Player.objects.get(pk=saved_player.pk)
+        assert player == saved_player
+
+    def test_create_player(self, saved_player, user):
+        player = saved_player
+        assert player.user == user
+        assert player.birth_date == '1990-01-01'
+        assert player.favorite_ball_game == BallGame.Soccer
```
yftacherzog, line 34: You can avoid this one. The test above is more powerful and doesn't require knowing the values for each of the fields.

player/tests.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def test_update_player(self, saved_player):
+        player = saved_player
```
yftacherzog, line 37: Add a line to assert that it was not basketball in the first place.
```suggestion
        player = saved_player
        assert updated_player.favorite_ball_game != BallGame.Basketball
```

yftacherzog (commented):
You're very close. Great work!

player/tests.py
Danielsio: ok, but what do you mean by encapsulating the user cration ? 
can you give me basic example maybe ?

yftacherzog: I'd think that you wouldn't have to create the user explicitly and that the creation of the player will also create the user. Need to make sure that is indeed a good idea and how to do it, so I don't want to do it now. We'll revisit it later when we have some more time.

player/tests.py
Danielsio: my thoughts exactly, nice.

player/tests.py
Danielsio: done

PR #29, opened 2023-04-16
Add PyTest and github workflow

### LINKED ISSUE #26: Need to Add PyTest and github workflow
Need to add:
1. PyTest
2. pytest.yml file to .github/workflows to automate pytest check.

IdanZimi:
- added PyTest to pipfile.
- added pytest.yml file which create a github workflow to test code file of every PR.
-  created dummy test in test_dummy.py and tests.py files to check github workflow, will be deleted later by next PR's. 
- added pytest.ini file.

closes #26

### DIFF
--- a/meet_balls_app/test_dummy.py
+++ b/meet_balls_app/test_dummy.py
@@ -3,4 +3,4 @@
 # Create your tests here.
 
 def test_dummy():
-    assert 1 + 2 == 3
+    assert 1 + 1 == 2
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,0 +1,16 @@
+name: PyTest
+on: pull_request
+jobs:
+  pytest:
+    name: Run the application's tests with PyTest
+    runs-on: ubuntu-20.04
+    container: fedora:34
+    steps:
+      - name: Install Pipenv and git
+        run: dnf install -y pipenv git
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup environment
+        run: pipenv sync --dev
+      - name: Run PyTest
+        run: pipenv run pytest -v 
--- a/Pipfile
+++ b/Pipfile
@@ -5,9 +5,11 @@ name = "pypi"
 
 [packages]
 django = "*"
+packaging = "==21.3"
 
 [dev-packages]
 flake8 = "*"
+pytest-django = "*"
 
 [requires]
 python_version = "3.9"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "bfea4da904cc0f6850b128b9d33331e8f854844bdd80edaa087f71f26391284c"
+            "sha256": "4106a2407c070493c5612a155998fa4e442c4fb92c2d59eb085ad6a6923c4413"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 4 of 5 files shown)


meet_balls_app/test_dummy.py
amitKaplansky, line 1: When we talked about it in slack yftach mentioned that the dummy test should be in tests.py file because it already exists. 
@yftacherzog Is it ok to leave it like that also?

IdanZimi: At first I put it right there at tests.py, but then the pytest checks didnt work.
after google it. it said pytest is searching for files with the prefix test_ or suffix _test to that's why I changed it.
if there is any other solution preferred I would love to hear

IdanZimi: I updated the PR. created the tests.py file.

amitKaplansky: I think you can add it to the pytest.yml file to configure a certain file or test directories.
I still don't have the answer, working on it.

amitKaplansky: I think you need to add this to the pytest.yml file and it might work. lmk (:
testpaths:
  - meet_balls_app/
python_files:
  - "test_*.py"
  - "tests.py"

IdanZimi: I think the solution (if needed at all) might be pytest.ini file and inside it add the following: 
`[pytest]
python_files = test_*.py *_test.py tests.py
`
is it worth changing? @yftacherzog

IdanZimi: > I think you need to add this to the pytest.yml file and it might work. lmk (: testpaths:
> 
> * meet_balls_app/
>   python_files:
> * "test_*.py"
> * "tests.py"

thanks! let's wait for a reply and move from there

amitCohen2: > I think the solution (if needed at all) might be pytest.ini file and inside it add the following: `[pytest] python_files = test_*.py *_test.py tests.py ` is it worth changing? @yftacherzog

I think it worth changing

Yuval-Vino: I think you should add `pytest.ini `, as you see without it you can't find files called tests.
this is a problem for us since Django generates for each app test.py if we leave it like this, you should always change the name of the file manually

IdanZimi: How `pytest.ini` related to the github workflow? 
shoudlnt I add some code to the yml file so it can detect files other than _test or test_?
Im confused. @Yuval-Vino

IdanZimi: anyway. I added the `pytest.ini` file. and also added dummy test to tests.py to ensure it check it too. you can see in the checks tab that it work now, it test both files `test_dummy.py` and `tests.py` as well. 
@Yuval-Vino  @amitKaplansky @amitCohen2

Yuval-Vino: @IdanZimi pytest.ini is for customizing some pytest behavior (such as what files to look for)  this should be placed in the root directory of the project
while the yml which is placed in the workflow

Yuval-Vino: @IdanZimi `pytest.ini` is for customizing Pytest behavior (such as which files to look for) and belongs in the project root directory. 
The YAML file used in GitHub workflows defines workflow steps and executes relevant terminal commands (in this case running pytest).

IdanZimi: Got it. So if I didnt tell nothing about which files to look for at the yml file, how github worflows know that it should also test the tests.py file? @Yuval-Vino

Yuval-Vino: basically, the yaml file just runs the `pipenv run pytest -v` command (and some more terminal commands)
the ini file is in charge of this command behavior

Danielsio (approved):
Isn't there a way to make pytest.ini file just search for every file contains the word test ?
other then that I think you have done a great job.

amitCohen2:
Looks like you fixed all the issues. Good job!

IdanZimi:
> Isnt there a way to make pytest.ini file just search for every file contains the word test ? other then that I think you have done a great job.

There is a way, but not sure if it's necessary. we already have a template to stick to:

-  *_test.py 
-   test_*.py 
-    tests.py

MaayanMashhadi (approved):

meet_balls_app/test_dummy.py
yftacherzog: I don't mind which convention you pick for supported file name formats as long as it allows having multiple test files. The file I provided was just a suggestion. I'm glad you have it figured out!

yftacherzog:
@IdanZimi I see now that the checks actually did run for this PR, contrary to what I thought is going to happen 😅

yftacherzog:
@IdanZimi make sure the commit message is following the guidelines on the contributing guide (the title should be `Add PyTest and github workflow` - capitalize, imperative mood, no period)

yftacherzog (changes requested):
Nice work on both the code and the discussion during code review! Fix the commit message and it can be merged.

yftacherzog (approved):

PR #30, opened 2023-04-16
Add GameEvent entity

### LINKED ISSUE #27: GameEvent entity needs to be added

MaayanMashhadi:
Create game event app with the starting files
Add GameEvent model to the game_event.models
Add GameEventConfig to the meet_balls/settings.py
Add tests for GameEvent entity

On this PR I work with @amitKaplansky 

Depends on #19
closes #27


game_event/models.py
```diff
@@ -0,0 +1,44 @@
...
+    Eight = 8
+    Nine = 9
+    Ten = 10
+
+
+class GameEvent(models.Model):
+    id = models.IntegerField(primary_key=True)
+    time = models.DateTimeField(null=False)
+    level_of_game = models.IntegerField(choices=Rating.choices, blank=True)
+    min_number_of_players = models.PositiveIntegerField()
```
Danielsio, line 35: Please consider removing the blank=True, also according to the django documentation null/blank Defaults are set to False, so you can remove the attribute completly if you like.

Danielsio: Maybe we can just set the game event level to be the lowest level possible for this case. That would achieve your goal.

yftacherzog: @amitKaplansky if this is something that blocking you, I suggest proceeding with the easiest approach to implement while opening an issue to have it figured out.

amitKaplansky: Should I add to the class a Meta class with unique_together = ('ball_responsible_player', 'court_id')? @yftacherzog @Yuval-Vino

yftacherzog: I don't know. Up to your design decisions.

I really really suggest that you proceed with the easiest way forward at this point and move to adding tests so the PR can be properly reviewed. You can (should) create an issue for any improvements you consider adding.

You now have 2 workdays to get this merged, and it's barely enough time for typical review cycles, and you still don't have the tests here, meaning the code is not complete, and you need to leave enough time to handle rebase issues etc.

At demo time, it will be much (much) more important to have everything merged (it must have proper tests to be merged) comparing to having some missing attributes.

game_event/models.py
```diff
@@ -0,0 +1,44 @@
...
+    max_number_of_players = models.PositiveIntegerField()
+    ball_responsible_player = models.ForeignKey(GameEventPlayer, on_delete=models.CASCADE)
```
yftacherzog, line 37: I think this part is going to be tricky. You have some sort of circular dependency here. 

You need the event to create the event player, and you need the event player to be the ball responsible. There's a way to work around this. It's mentioned here (with oddly similar case): https://stackoverflow.com/questions/8466726/django-circular-model-reference

If you don't get it figured out easily enough, I suggest having this reference a player rather than an event player for now, and open an issue to figure it out later.

MaayanMashhadi: > I think this part is going to be tricky. You have some sort of circular dependency here.
> 
> You need the event to create the event player, and you need the event player to be the ball responsible. There's a way to work around this. It's mentioned here (with oddly similar case): https://stackoverflow.com/questions/8466726/django-circular-model-reference
> 
> If you don't get it figured out easily enough, I suggest having this reference a player rather than an event player for now, and open an issue to figure it out later.

Thank you for this comment, we will see what we do about it and let you know

game_event/models.py
```diff
@@ -0,0 +1,44 @@
...
+    court_id = models.ForeignKey(Court, on_delete=models.CASCADE)
+    ball_game = models.CharField(
+        max_length=100,
```
SchSeba, line 40: small comment I think you can make the length lower here

game_event/models.py
yftacherzog: I have another idea for avoiding this circular dependency. We can have a ball_responsible boolean on the EventPlayer model. It will be False by default, and True for the ones who are ball responsible.

This way, we only have one way dependency.

game_event/models.py
MaayanMashhadi: > small comment I think you can make the length lower here

done

game_event/models.py
MaayanMashhadi: > I have another idea for avoiding this circular dependency. We remove this field here, and add a ball_responsible boolean on the EventPlayer model. It will be False by default, and True for the ones who are ball responsible.
> 
> This way, we only have one way dependency.

So, just to make sure I understand: on GameEvent we don't need the ball responsible, the EventPlayer entity
will be responsible about this field?

SchSeba: and you can add a validation that the player with the ball is one from the players list

yftacherzog:
@MaayanMashhadi @amitKaplansky if I understand correctly, you now have no dependency on anything yet to be merged. I suggest you rebase the PR so you're left with only the relevant content and make it ready to be merged. If you're running into trouble, ping me early on tomorrow so we can have it fixed.

MaayanMashhadi:
> @MaayanMashhadi @amitKaplansky if I understand correctly, you now have no dependency on anything yet to be merged. I suggest you rebase the PR so you're left with only the relevant content and make it ready to be merged. If you're running into trouble, ping me early on tomorrow so we can have it fixed.

done as requested

yftacherzog:
> > @MaayanMashhadi @amitKaplansky if I understand correctly, you now have no dependency on anything yet to be merged. I suggest you rebase the PR so you're left with only the relevant content and make it ready to be merged. If you're running into trouble, ping me early on tomorrow so we can have it fixed.
> 
> 
> 
> done as requested

I still see 2 commits. Also make sure the pr description is up to date. Eg no mentions of stubs etc.

amitKaplansky:
> > > @MaayanMashhadi @amitKaplansky if I understand correctly, you now have no dependency on anything yet to be merged. I suggest you rebase the PR so you're left with only the relevant content and make it ready to be merged. If you're running into trouble, ping me early on tomorrow so we can have it fixed.
> > 
> > 
> > done as requested
> 
> I still see 2 commits. Also make sure the pr description is up to date. Eg no mentions of stubs etc.

> > > @MaayanMashhadi @amitKaplansky if I understand correctly, you now have no dependency on anything yet to be merged. I suggest you rebase the PR so you're left with only the relevant content and make it ready to be merged. If you're running into trouble, ping me early on tomorrow so we can have it fixed.
> > 
> > 
> > done as requested
> 
> I still see 2 commits. Also make sure the pr description is up to date. Eg no mentions of stubs etc.

The second commit is because we needed GameEventPlayer entity should we remove it?

yftacherzog:
> > > > @MaayanMashhadi @amitKaplansky if I understand correctly, you now have no dependency on anything yet to be merged. I suggest you rebase the PR so you're left with only the relevant content and make it ready to be merged. If you're running into trouble, ping me early on tomorrow so we can have it fixed.
> 
> > > 
> 
> > > 
> 
> > > done as requested
> 
> > 
> 
> > I still see 2 commits. Also make sure the pr description is up to date. Eg no mentions of stubs etc.
> 
> 
> 
> > > > @MaayanMashhadi @amitKaplansky if I understand correctly, you now have no dependency on anything yet to be merged. I suggest you rebase the PR so you're left with only the relevant content and make it ready to be merged. If you're running into trouble, ping me early on tomorrow so we can have it fixed.
> 
> > > 
> 
> > > 
> 
> > > done as requested
> 
> > 
> 
> > I still see 2 commits. Also make sure the pr description is up to date. Eg no mentions of stubs etc.
> 
> 
> 
> The second commit is because we needed GameEventPlayer entity should we remove it?

Event player should be in a pr of its own. If I see it correctly, event has no unnerved dependencies. Is that right?

You need to leave inside this pr only the commit that should be merged as part of this pr. 

The event player commit will be part of the event player pr only.

MaayanMashhadi:
After mergging the GameEventPlayer into main, the test will pass.
The test in game_event app depends on game_event_player app

meetballs/settings.py
```diff
@@ -40,7 +40,9 @@
     'meet_balls_app.apps.MeetBallsAppConfig',
     'player_rating.apps.PlayerRatingConfig',
     'court.apps.CourtConfig',
-    'player.apps.PlayerConfig'
+    'player.apps.PlayerConfig',
+    'game_event.apps.GameEventConfig',
+    'game_event_player.apps.GameEventPlayerConfig'
 ]
```
Danielsio, line 46: this PR, doesn't need line 45 here, so if you remove it it would solve your test fail, am i wrong ?

if i am right, that would make your PR ez to merge.

Danielsio:
> After mergging the GameEventPlayer into main, the test will pass. The test in game_event app depends on game_event_player app

Hi Maayan I went through your tests, didn't see any test that requires game_event_player, but if you have one anywhere, i think it's better to remove it temporarily, create issue to add it later after dependency is merged, since it's a valid test it shouldn't take long to review again and quickly merge it to complete your work. 

assuming of course this PR suppose to be the next one thats geting merged.

yftacherzog:
> After mergging the GameEventPlayer into main, the test will pass.
> 
> The test in game_event app depends on game_event_player app

I think the dependency is the other way around. The event player needs the event (it references it). The event doesn't reference the event player anymore it should not depend on it.

MaayanMashhadi:
> > After mergging the GameEventPlayer into main, the test will pass.
> > The test in game_event app depends on game_event_player app
> 
> I think the dependency is the other way around. The event player needs the event (it references it). The event doesn't reference the event player anymore it should not depend on it.



> > After mergging the GameEventPlayer into main, the test will pass.
> > The test in game_event app depends on game_event_player app
> 
> I think the dependency is the other way around. The event player needs the event (it references it). The event doesn't reference the event player anymore it should not depend on it.

You're right. But I don't know the what is the problem... It says to me that
the module don't found but I didn't use it. I try to remove the initial_0001, and do makemigration
but now this isn't work on the same reason. it writes tome the next error:
ModuleNotFoundError: No module named 'game_event_player'
I didn't see any use of this module on this branch.

meetballs/settings.py
MaayanMashhadi: > this PR, doesn't need line 45 here, so if you remove it it would solve your test fail, am i wrong ?
> 
> if i am right, that would make your PR ez to merge.

Amazing! you're right, it fixes the problem
Thank you

meet_balls_app/models.py
yftacherzog, line 1: We try to avoid by-the-way changes in PRs. We limit the scope of changes to the things relevant to the PR. It reduces surprises by making the PR intention more transparent and it makes code review easier.

In this case, the PR is not related to this files, so remove the changes made to this file from the PR.

game_event/tests.py
```diff
@@ -0,0 +1,76 @@
...
+    def test_update_game_event(self, saved_game_event):
+        game_event = saved_game_event
+        assert saved_game_event.ball_game != BallGame.Volleyball
+        game_event.ball_game = BallGame.Volleyball
+        game_event.save()
+        updated_game_event = GameEvent.objects.get(pk=game_event.pk)
+        assert updated_game_event.ball_game == BallGame.Volleyball
+
+    def test_delete_game_event(self, saved_game_event):
+        game_event = saved_game_event
```
yftacherzog, line 44: You're not actually using `game_event` after the removal.
I'm not certain if you need it or if it is of any benefit to use it.

This is how I would do it (slightly different from what we had on the previous modules):
```
    def test_delete_game_event(self, saved_game_event):
        pk = saved_game_event.pk
        saved_game_event.delete()
        with pytest.raises(GameEvent.DoesNotExist):
            GameEvent.objects.get(pk=pk)
```

yftacherzog (changes requested):
almost there

yftacherzog (approved):

IdanZimi (approved):

game_event/models.py
```diff
@@ -0,0 +1,30 @@
+from django.db import models
+from django.core.validators import MinLengthValidator
+from court.models import Court
+from player_rating.models import Rating
+
+
+class BallGame(models.TextChoices):
```
Yuval-Vino, line 7: Isn't it already defined in the Player model?
you can just import it

yftacherzog: @MaayanMashhadi @amitKaplansky I merged before seeing this comment. Please check it and open an issue accordingly.

PR #31, opened 2023-04-16
Add GameEventPlayer entity

### LINKED ISSUE #28: GameEventPlayer entity needs to be added

MaayanMashhadi:
Create game_event_player app with the starting files
Add GameEventPlayer model to the game_event_player.models
Add GameEventPlayerConfig to the meet_balls/settings.py
Add tests for GameEventPlayer entity

On this PR I work with @amitKaplansky 

Depends on #30 #21 
closes #28


amitKaplansky:
waiting for your code review @IdanZimi @ZvikaNaorCohen @amitCohen2 @Danielsio @OhadSeg

game_event_player/models.py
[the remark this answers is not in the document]
amitKaplansky: In Django you only have the possibility to have only one PK, in our case we have two. As @Yuval-Vino requested we added the unique_together, in this why they depend on each other.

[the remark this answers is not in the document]
Yuval-Vino: The reason is, if we do not do it, it will be possible for a player to enter multiple times to the Event since the PK is GameEventPlayer.id and generated automatically (not looking at userID or eventID) @OhadSeg

game_event_player/models.py
```diff
@@ -0,0 +1,11 @@
+from django.db import models
+from game_event.models import GameEvent
+from player.models import Player
+
+
+class GameEventPlayer(models.Model):
+    game_event_id = models.ForeignKey(GameEvent, on_delete=models.CASCADE)
+    user_id = models.ForeignKey(Player, on_delete=models.CASCADE)
```
SchSeba, line 8: Hi sorry to by a bit late to the party.

question why we don't use https://docs.djangoproject.com/en/4.2/ref/models/fields/#manytomanyfield in the game_event model. 

django will create the connection table for you

game_event/admin.py
ZvikaNaorCohen, line 1: In my opinion we should register all entities to admin.py file, I can't find it here. Is this entity expected to be registered? For example, in player's entity we used: 
image

game_event_player/admin.py
```diff
@@ -0,0 +1 @@
+# from django.contrib import admin
```
Danielsio, line 1: @ZvikaNaorCohen That's the place you meant to address your last comment I assume. 

anyway as naor said, GameEventPlayer entity Should be registered to admin here.

yftacherzog:
> test_create_game_event_player_with_empty_game_event() is failing because the class GameEvent is empty. In reality the class isn't empty so the test won't fail.

@MaayanMashhadi @amitKaplansky you can rebase on top of the branch of the event PR. I don't expect major changes there at this point, and it will allow you to finalize the work on this one.

game_event_player/tests.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def test_create_game_event_player_with_empty_player(self, saved_game_event):
+        with pytest.raises(IntegrityError):
+            GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=Player.objects.create(),
+                                           ball_responsible=False)
+
+    def test_create_game_event_player_with_empty_game_event(self, saved_player):
+        with pytest.raises(IntegrityError):
+            GameEventPlayer.objects.create(game_event_id=GameEvent.objects.create(), player_id=saved_player,
+                                           ball_responsible=False).full_clean()
```
Yuval-Vino, line 71: does `full_clean()` necessary here?
I think you should get an error even without it

MaayanMashhadi: > does `full_clean()` necessary here? I think you should get an error even without it

You're right, the test passes without full_clean()

game_event_player/tests.py
```diff
@@ -0,0 +1,94 @@
...
+@pytest.mark.django_db
+def user():
+    return get_user_model().objects.create_user(
+        username='testuser',
+        password='testpass'
+    )
+
+
+@pytest.fixture
+@pytest.mark.django_db
```
Yuval-Vino, line 23: i think this is not necessary `@pytest.mark.django_db`

game_event_player/tests.py
```diff
@@ -0,0 +1,94 @@
...
+
+@pytest.fixture
+def saved_court():
+    court = Court.objects.create(x=Decimal('11'), y=Decimal('22'),
+                                 city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
+    return court
+
+
+@pytest.fixture
+@pytest.mark.django_db
```
Yuval-Vino, line 41: same here

game_event_player/tests.py
```diff
@@ -0,0 +1,94 @@
...
+def saved_game_event(saved_court):
+    game_event = GameEvent.objects.create(id=2, time=datetime.today(), level_of_game=3,
+                                          min_number_of_players=2, max_number_of_players=5, court_id=saved_court,
+                                          ball_game='Basketball')
+    return game_event
+
+
+@pytest.fixture
+@pytest.mark.django_db
```
Yuval-Vino, line 50: same here

game_event_player/tests.py
MaayanMashhadi: > i think this is not necessary `@pytest.mark.django_db`, should only be above the Test Class

The fixtures functions are not inside the class. If you want me to remove the @pytest.mark.django_db
it means the functions should be inside the class. Is that what you want?
After the confitest.py PR is merged we will need to add it all those files meaning fixtures wont be inside of a class.

Yuval-Vino: @MaayanMashhadi mark.django_db is used to mark a test function as requiring the database
the fixture is for providing data for tests.

there is no reason to put the fixtures inside the class or under`@pytest.mark.django_db`

try removing it and see what is the outcome

MaayanMashhadi: done

game_event_player/tests.py
```diff
@@ -0,0 +1,90 @@
...
+    return GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player,
+                                          ball_responsible=False)
+
+
+@pytest.mark.django_db
+class TestGameEventPlayerModel:
+
+    def test_game_event_player_creation(self, saved_game_event, saved_player):
+        game_event_player = GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player,
+                                                           ball_responsible=False)
```
yftacherzog, line 57: Can you call `full_clean` here to show that it passes validation when the inputs are good.

MaayanMashhadi: > Can you call `full_clean` here to show that it passes validation when the inputs are good?

Do I need to add the full_clean() after the assert commands?

game_event_player/tests.py
```diff
@@ -0,0 +1,90 @@
...
+        assert game_event_player.game_event_id == saved_game_event
+        assert game_event_player.player_id == saved_player
+        assert game_event_player.ball_responsible is False
+
+    def test_unique_together_constraint(self, saved_game_event, saved_player):
+        GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player, ball_responsible=False)
+        with pytest.raises(IntegrityError):
+            GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player,
+                                           ball_responsible=False)
```
yftacherzog, line 66: I'd add a positive test case in which you create 3 players for one event and also add 2 of them to another event, and then do some filtering: seeing all events for a specific players, and all the players of a specific event.

game_event_player/tests.py
yftacherzog: I'd suggest maybe to drop the assignment to `ball_responsible` so we can test the default value.

yftacherzog (commented):
almost there :)

game_event_player/tests.py
yftacherzog: > > Can you call `full_clean` here to show that it passes validation when the inputs are good?
> 
> Do I need to add the full_clean() after the assert commands?

```suggestion
        game_event_player = GameEventPlayer.objects.create(
                game_event_id=saved_game_event,
                 player_id=saved_player,
        ).full_clean()
```

MaayanMashhadi: > > > Can you call `full_clean` here to show that it passes validation when the inputs are good?
> > 
> > 
> > Do I need to add the full_clean() after the assert commands?

When I call "full_clean()" function, if all checks pass, it returns 'None' (I read on chat gpt)
and we use the game_event_player after. Do I need to add "if is None" statement after do full_clean()?

yftacherzog: ok. So I suppose you'd need to add another test in which you will call full_clean and assert it's not throwing anything. I think we did similar stuff on the other PRs. See how it's done there.

MaayanMashhadi: done

game_event_player/tests.py
```diff
@@ -0,0 +1,172 @@
...
+        assert game_event_player.game_event_id == saved_game_event
+        assert game_event_player.player_id == saved_player
+        assert game_event_player.ball_responsible is False
+
+    def test_game_event_player_ball_responsible_default_value(self, saved_game_event, saved_player):
+        game_event_player = GameEventPlayer.objects.create(game_event_id=saved_game_event,
+                                                           player_id=saved_player,).full_clean()
+        assert game_event_player is None
+
+    def test_filter_players_by_event(self, saved_game_event, saved_game_event2, players):
```
Yuval-Vino, line 129: This test may not be entirely self-explanatory. To improve readability, consider dividing it into three parts and encapsulating each part within a separate function:

1. function for creating GameEventPlayers.
2. a function that will test whether a player is in the expected events.
3. a function that will test whether an event has the expected players.

Breaking down the test into smaller, focused functions will make it easier to read and understand.

game_event_player/tests.py
```diff
@@ -0,0 +1,172 @@
...
+        game_event_player = GameEventPlayer.objects.create(game_event_id=saved_game_event,
+                                                           player_id=saved_player,)
+        assert game_event_player.game_event_id == saved_game_event
+        assert game_event_player.player_id == saved_player
+        assert game_event_player.ball_responsible is False
+
+    def test_game_event_player_ball_responsible_default_value(self, saved_game_event, saved_player):
+        game_event_player = GameEventPlayer.objects.create(game_event_id=saved_game_event,
+                                                           player_id=saved_player,).full_clean()
+        assert game_event_player is None
```
Yuval-Vino, line 127: i think what you want to test here is this: 
```suggestion
      def test_game_event_player_ball_responsible_default_value(self, saved_game_event, saved_player):
      GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player,).full_clean()
```
we just want to make sure full_clean doesn't raise any errors,

yftacherzog: I agree with @Yuval-Vino here

game_event_player/tests.py
yftacherzog: It's not the prettiest of tests, but we're already over the time we agreed we'll merge stuff. I suggest opening an issue for cleaning this up.

game_event_player/tests.py
yftacherzog: I think we had a miscommunication here. What @Yuval-Vino suggested on the other thread is the way to go.

game_event_player/tests.py
```diff
@@ -0,0 +1,172 @@
...
+    return GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player,
+                                          ball_responsible=False)
+
+
+@pytest.fixture
+def create_game_event_players(saved_game_event, saved_players, size):
+    players = []
+    for i in range(size):
+        players.append(GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_players[i],
+                                                      ball_responsible=False))
```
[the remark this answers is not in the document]
MaayanMashhadi: > Are you actually using this? In any case, you can do it in more pythonic way, and I don't think you need the size:

It does to me a syntax error. I will try to figure out

game_event_player/tests.py
amitKaplansky: Opened an issue related to this #52

game_event_player/tests.py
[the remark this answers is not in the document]
MaayanMashhadi: > Are you actually using this? In any case, you can do it in more pythonic way, and I don't think you need the size:

done

game_event_player/tests.py
MaayanMashhadi: > i think what you want to test here is this:
> 
> we just want to make sure full_clean doesn't raise any errors,

done

game_event_player/tests.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+@pytest.fixture
+def saved_game_event_player(saved_game_event, saved_player):
+    return GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player,
+                                          ball_responsible=False)
+
+
+@pytest.fixture
+def create_game_event_players(saved_game_event, saved_players):
```
yftacherzog, line 106: If it's not being used, then just delete it.

game_event_player/tests.py
```diff
@@ -0,0 +1,168 @@
...
+    players = [GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=player)
+               for player in saved_players]
+    return players
+
+
+@pytest.mark.django_db
+class TestGameEventPlayerModel:
+
+    def test_game_event_player_ball_responsible_default_value(self, saved_game_event, saved_player):
+        GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player, ).full_clean()
```
yftacherzog, line 116: 🤦 
Looking at the title, now we're not testing what were supposed to test here 
```suggestion
    def test_game_event_player_ball_responsible_default_value(self, saved_game_event, saved_player):
        instance = GameEventPlayer.objects.create(game_event_id=saved_game_event, player_id=saved_player)
        instance.full_clean()
        assert instance.ball_responsible is False
```

game_event_player/tests.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def test_game_event_player_creation(self, saved_game_event, saved_player):
+        game_event_player = GameEventPlayer.objects.create(game_event_id=saved_game_event,
+                                                           player_id=saved_player,)
+        assert game_event_player.game_event_id == saved_game_event
+        assert game_event_player.player_id == saved_player
+        assert game_event_player.ball_responsible is False
```
yftacherzog, line 123: You can make this change and then you can delete the test case above. Its name doesn't make sense and this line will also make it redundant.
```suggestion
    def test_game_event_player_creation(self, saved_game_event, saved_player):
        game_event_player = GameEventPlayer.objects.create(game_event_id=saved_game_event,
                                                           player_id=saved_player,)
        game_event_player.full_clean()
        assert game_event_player.game_event_id == saved_game_event
        assert game_event_player.player_id == saved_player
        assert game_event_player.ball_responsible is False
```

game_event_player/tests.py
yftacherzog, line 116: The name of the test doesn't make sense and I think it can be made redundant by the addition suggested to the test below.

PR #33, opened 2023-04-16
Add player-rating class

### LINKED ISSUE #32: Need to Add Player-Rating Entity

Danielsio:
Add PlayerRating to the player/models
Declare Rating as IntegerChoices
Add tests for PlayerRating  at player/tests

Depends On #21 

closes #32


Danielsio:
Please review only the second commit.

Danielsio:
I decided to create another app for player_rating rather then just add it to the existing player app.

player_rating/tests.py
yftacherzog, line 1: I think you can avoid the word `fixture` in the fixture names.

player_rating/tests.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+        return player_rating_fixture
+
+    def test_get_player_rating(self, player_rating_fixture):
+        player_rating_from_db = PlayerRating.objects.get(
+            pk=player_rating_fixture.pk)
+        assert player_rating_from_db == player_rating_fixture
+
+    def test_update_player_rating(self, player_rating_fixture):
+        player_rating_fixture.rating = Rating.Seven
```
yftacherzog, line 43: Add an assertion that it was not seven before the change to make the test extra clear and safe.

player_rating/tests.py
```diff
@@ -0,0 +1,103 @@
...
+                ball_game=BallGame.Soccer,
+                user_id=player_rating_fixture.user_id,
+                rating=0).full_clean()
+
+    def test_delete_player_deletes_player_ratings(self, player_fixture, player_rating_fixture):
+        assert Player.objects.count() == 1
+        assert PlayerRating.objects.count() == 1
+        player_fixture.delete()
+        assert Player.objects.count() == 0
+        assert PlayerRating.objects.count() == 0
```
yftacherzog, line 89: Nice!

yftacherzog (changes requested):
Nicely done! Almost there.

player_rating/tests.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def test_create_player_rating_invalid_ball_game(self, player_rating_fixture):
+        with pytest.raises(ValidationError):
+            PlayerRating.objects.create(
+                ball_game='Invalid Ball Game',
+                user_id=player_rating_fixture.user_id,
+                rating=Rating.Five).full_clean()
+
+    def test_create_player_rating_invalid_user_id(self):
+        with pytest.raises(IntegrityError):
```
yftacherzog, line 65: Can you check if this is raised by the call for full_clean or earlier? Try removing the call for full_clean and see if it's still passes. If it does, make sure you remove this from all tests that expects `IntegrityError`s.

Danielsio: You are correct, the test indeed passes without the full_clean()

player_rating/tests.py
Danielsio: Done.

player_rating/tests.py
Danielsio: Done, it was originaly named player_from_fixture that become player_fixture

to make it easy for the reader to understand from where the instance is coming from.

yftacherzog (approved):

Yuval-Vino (approved):
Well done, Daniel!
I noticed that you defined `pytest.fixture` here, which has already been defined in Player tests. Perhaps there is a way to consolidate all fixtures into a single Python file and use them from there. This will be beneficial as other model tests will also require the creation of a user and a player.
I'm unsure if this is applicable to the current pull request, but it could be useful for future reference.

PR #47, opened 2023-04-20
Add Create Method for Player.

### LINKED ISSUE #36: Add a  function that will encapsulate the creation of the user when creating the player.

Danielsio:
### Desired Outcome

Player should'nt be created without a user in it.  

### Implemented Changes

I (with some help) Added static method "create" for the Player class
method recives parameters for user creation, and the rest of the player fields

all tests regarding creation of a player have been changed - you should just make sure there are no redandent tests there

### Connected Issue/Story

Resolves #36

#### Test coverage

- [x] This PR includes new unit tests to go with the code changes

### DIFF
--- a/player/tests.py
+++ b/player/tests.py
@@ -1,5 +1,7 @@
 from .models import Player, BallGame
 from django.core.exceptions import ValidationError
+from django.contrib.auth.models import User
+from django.db import IntegrityError
 import pytest
 
 
@@ -11,12 +13,30 @@ class TestPlayerModel:
         assert player_from_db == player
 
     def test_update_player(self, player):
-        player_copy = player
-        assert player_copy.favorite_ball_game != BallGame.Basketball
-        player_copy.favorite_ball_game = BallGame.Basketball
-        player_copy.save()
-        updated_player = Player.objects.get(pk=player_copy.pk)
+        assert player.favorite_ball_game != BallGame.Basketball
+        player.favorite_ball_game = BallGame.Basketball
+        player.save()
+        updated_player = Player.objects.get(pk=player.pk)
         assert updated_player.favorite_ball_game == BallGame.Basketball
+
+    def test_create_player_with_new_user(self):
+        player = Player.create(username='daniel',
+                               password='password',
+                               birth_date='1990-01-01',
+                               favorite_ball_game=BallGame.Soccer)
+        assert isinstance(player.user, User)
+        assert player.user.username == 'daniel'
+        assert player.user.check_password('password')
+
+    def test_create_player_with_existing_username(self):
+        user = User.objects.create_user(username='daniel', password='password')
+        user.save()
+        with pytest.raises(IntegrityError):
+            Player.create(
+                username=user.username,
+                password=user.password,
+                birth_date='1990-01-01',
+                favorite_ball_game=BallGame.Soccer)
 
     def test_delete_player(self, player):
         player_copy = player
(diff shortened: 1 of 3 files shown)


yftacherzog (commented):
I'd expect to see test cases for validating the underlying user, both when the user existed before and when it didn't.

Danielsio:
> I'd expect to see test cases for validating the underlying user, both when the user existed before and when it didn't.

so, if i understand correctly, i still need the user fixture.. for the test case that uses the existing_user.

Danielsio:
> > > I'd expect to see test cases for validating the underlying user, both when the user existed before and when it didn't.
> > 
> > 
> > so, if i understand correctly, i still need the user fixture.. for the test case that uses the existing_user.
> 
> Sure. You can use a fixture for that.

Done.

problem now is when i change the implementation of save method from :

def save(self, *args, **kwargs):
        if not self.user_id:
            user = User.objects.create_user(
                username='default_username',
                password='default_password')
            self.user = user
        super().save(*args, **kwargs)
        
        to this :
        
        def save(self, *args, **kwargs):
        if not self.user:
            try:
                user = User.objects.create_user(
                    username='default_username',
                    password='default_password')
                self.user = user
                self.user.save()
            except IntegrityError:
                # User already exists, do nothing
                pass
        super().save(*args, **kwargs)
        
        like you seggested, somehow all my test failing with Error:
        Player has no user.
        
        cant understand why..

yftacherzog:
> > > > I'd expect to see test cases for validating the underlying user, both when the user existed before and when it didn't.
> > > 
> > > 
> > > so, if i understand correctly, i still need the user fixture.. for the test case that uses the existing_user.
> > 
> > 
> > Sure. You can use a fixture for that.
> 
> Done.
> 
> problem now is when i change the implementation of save method from :
> 
> ```
>     def save(self, *args, **kwargs):
>     if not self.user_id:
>         user = User.objects.create_user(
>             username='default_username',
>             password='default_password')
>         self.user = user
>     super().save(*args, **kwargs)
>     
>     to this :
>     
>     def save(self, *args, **kwargs):
>     if not self.user:
>         try:
>             user = User.objects.create_user(
>                 username='default_username',
>                 password='default_password')
>             self.user = user
>             self.user.save()
>         except IntegrityError:
>             # User already exists, do nothing
>             pass
>     super().save(*args, **kwargs)
>     
>     like you seggested, somehow all my test failing with Error:
>     Player has no user.
>     
>     cant understand why.. 
> ```

Inside the except, you still need to set `self.user`. Might be that you fail on that.

Danielsio:
> > > > > I'd expect to see test cases for validating the underlying user, both when the user existed before and when it didn't.
> > > > 
> > > > 
> > > > so, if i understand correctly, i still need the user fixture.. for the test case that uses the existing_user.
> > > 
> > > 
> > > Sure. You can use a fixture for that.
> > 
> > 
> > Done.
> > problem now is when i change the implementation of save method from :
> > ```
> >     def save(self, *args, **kwargs):
> >     if not self.user_id:
> >         user = User.objects.create_user(
> >             username='default_username',
> >             password='default_password')
> >         self.user = user
> >     super().save(*args, **kwargs)
> >     
> >     to this :
> >     
> >     def save(self, *args, **kwargs):
> >     if not self.user:
> >         try:
> >             user = User.objects.create_user(
> >                 username='default_username',
> >                 password='default_password')
> >             self.user = user
> >             self.user.save()
> >         except IntegrityError:
> >             # User already exists, do nothing
> >             pass
> >     super().save(*args, **kwargs)
> >     
> >     like you seggested, somehow all my test failing with Error:
> >     Player has no user.
> >     
> >     cant understand why.. 
> > ```
> 
> Inside the except, you still need to set `self.user`. Might be that you fail on that.

I tried.. didnt work out the way i expected, any suggestions what am i doing worng ?

yftacherzog:
> > > > > > I'd expect to see test cases for validating the underlying user, both when the user existed before and when it didn't.
> > > > > 
> > > > > 
> > > > > so, if i understand correctly, i still need the user fixture.. for the test case that uses the existing_user.
> > > > 
> > > > 
> > > > Sure. You can use a fixture for that.
> > > 
> > > 
> > > Done.
> > > problem now is when i change the implementation of save method from :
> > > ```
> > >     def save(self, *args, **kwargs):
> > >     if not self.user_id:
> > >         user = User.objects.create_user(
> > >             username='default_username',
> > >             password='default_password')
> > >         self.user = user
> > >     super().save(*args, **kwargs)
> > >     
> > >     to this :
> > >     
> > >     def save(self, *args, **kwargs):
> > >     if not self.user:
> > >         try:
> > >             user = User.objects.create_user(
> > >                 username='default_username',
> > >                 password='default_password')
> > >             self.user = user
> > >             self.user.save()
> > >         except IntegrityError:
> > >             # User already exists, do nothing
> > >             pass
> > >     super().save(*args, **kwargs)
> > >     
> > >     like you seggested, somehow all my test failing with Error:
> > >     Player has no user.
> > >     
> > >     cant understand why.. 
> > > ```
> > 
> > 
> > Inside the except, you still need to set `self.user`. Might be that you fail on that.
> 
> I tried.. didnt work out the way i expected, any suggestions what am i doing worng ?

We can have a look together next week.

player/tests.py
```diff
@@ -1,28 +1,31 @@
...
+        player = Player.create(
+            username="daniel",
+            password="password",
             birth_date='1990-01-01',
-            favorite_ball_game=BallGame.Soccer
-        )
+            favorite_ball_game=BallGame.Soccer)
         return player
 
+    def test_create_player_with_new_user(self):
```
yftacherzog, line 20: What will happen if the user already exist? It's up to you to decide what's the expected behavior, but you need a test case that shows it's working as expected.

player/tests.py
```diff
@@ -1,28 +1,41 @@
...
+        player = Player.create(username='daniel',
+                               password='password',
+                               birth_date='1990-01-01',
+                               favorite_ball_game=BallGame.Soccer)
+        assert isinstance(player.user, User)
+        assert player.user.username == 'daniel'
+        assert player.user.check_password('password')
+
+    def test_create_player_with_existing_user(self):
```
Yuval-Vino, line 29: I think a better name will be `test_create_player_with_existing_username`, since you are not passing the `user` variable, and the creating is failed because in `Player.create` method we trying to create a new User with an existing username

Danielsio: I agree.

Yuval-Vino (approved):

player/tests.py
Danielsio: If there is a user with a username equal to the username sent to the Player.create,  IntegrityError will raised.

MaayanMashhadi:
Hi Daniel,
I have just one question: why did you do this method?
for what reason? just for understanding
Besides, very good work!

player/tests.py
```diff
@@ -47,43 +60,50 @@ def test_delete_user_deletes_player(self, saved_player):
...
+                password="password",
                 birth_date='1990-01-01',
                 favorite_ball_game='').full_clean()
 
-    def test_create_player_with_null_favorite_ball_game(self, user):
-        with pytest.raises(ValidationError):
-            Player.objects.create(
-                user=user,
+    def test_create_player_with_null_favorite_ball_game(self):
+        with pytest.raises(IntegrityError):
```
MaayanMashhadi, line 103: As I see, the purpose of the test is to raise error when you put
None value on favorite_ball_game field.
You don't need the ValidationError here?
As I know, ValidtionError is for that case or maybe I wrong?

PR #54, opened 2023-04-23
Removing BallGame class from game_event models

### LINKED ISSUE #49: Remove BallGame class from game_event models.py
removing the class BallGame from game_event models, because it is already in Player class.

MaayanMashhadi:
Removing BallGame class from game_event models
alse changing imports from tests file
closes #49


Danielsio (approved):
Nice Work Maayan.

Yuval-Vino (approved):

PR #65, opened 2023-04-30
Add base.html to meet_balls_app/templates

### LINKED ISSUE #64: Add Base.html file for navbar
## Expected Behavior

Application templates should have a base.html template file so they would extends it.

## Current Behavior

There is no templates at all.

## Acceptance Criteria

Add the base.html template to the meet_balls_app/templates

Danielsio:
### Desired Outcome

create a simple base.html file for other future templates to inherit 

### Implemented Changes

add base.html file to meet_balls_app/templates

in the file you will find links to bootstrap cdn and super basic Navbar implementation.

### Connected Issue/Story

Resolves #64

### DIFF
--- a/meet_balls_app/templates/base.html
+++ b/meet_balls_app/templates/base.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>MeetBalls</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link href="/static/css/styles.css" rel="stylesheet" />
+    <link
+      href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/css/bootstrap.min.css"
+      rel="stylesheet"
+    />
+    <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/js/bootstrap.bundle.min.js"></script>
+  </head>
+  <body>
+    <nav class="navbar navbar-expand-md navbar bg-warning">
+      <div class="container-fluid">
+        <a class="navbar-brand" href="/">MeetBalls</a>
+        <button
+          class="navbar-toggler"
+          type="button"
+          data-bs-toggle="collapse"
+          data-bs-target="#navbarNav"
+          aria-controls="navbarNav"
+          aria-expanded="false"
+          aria-label="Toggle navigation"
+        >
+          <span class="navbar-toggler-icon"></span>
+        </button>
+        <div class="collapse navbar-collapse" id="navbarNav">
+          <ul class="navbar-nav ms-auto">
+            <li class="nav-item">
+              <a class="nav-link" href="/game-events">Games</a>
+            </li>
+            <li class="nav-item">
+              <a class="nav-link" href="/about">About</a>
+            </li>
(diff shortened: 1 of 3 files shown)


Danielsio:
In addition to the description, I will explain that the changes I made were only to create the base file and change the home.html existing file so that it inherits from the base, thereby checking that it really works.

one more thing i would like to know if tests are required for this kind of PR, if so what should i test for ?

meet_balls_app/templates/base.html
```diff
@@ -0,0 +1,52 @@
...
+          aria-expanded="false"
+          aria-label="Toggle navigation"
+        >
+          <span class="navbar-toggler-icon"></span>
+        </button>
+        <div class="collapse navbar-collapse" id="navbarNav">
+          <ul class="navbar-nav ms-auto">
+            <li class="nav-item">
+              <a class="nav-link" href="/game-events">Games</a>
+            </li>
```
Danielsio, line 32: The href will redirect the user to /game-events url which isn't defined right now.
It's on perpose so when game_events template PR will be pushed it will be defined.

if it's problematic the way it is right now lem me know, i will change it to "#" so it wont go crazy on us

IdanZimi (commented):
LGTM. Would like to know as you said if there's a way to test this code

yftacherzog:
> In addition to the description, I will explain that the changes I made were only to create the base file and change the home.html existing file so that it inherits from the base, thereby checking that it really works.
> 
> one more thing i would like to know if tests are required for this kind of PR, if so what should i test for ?

What you can do is to add a test case for a `get` call on the main page and assert on the templates being called (as done on the class this week). I don't know however if this will list the base template or not. It will be good to have it anyhow as an example for testing templates usage.

yftacherzog (approved):

meet_balls_app/templates/base.html
```diff
@@ -0,0 +1,52 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>MeetBalls</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link href="/static/css/styles.css" rel="stylesheet" />
```
SchSeba, line 6: this file doesn't exist in the repo right?

meet_balls_app/tests.py
```diff
@@ -1,3 +1,6 @@
-# from django.test import TestCase
+class TestUi:
 
-# Create your tests here.
+    def test_home_page_extends_base(self, client):
```
SchSeba, line 3: can you please validate here that you see the login and create another test where you have a user that is already after login and check you see the logout please

meet_balls_app/templates/base.html
Danielsio: That's correct, first one to add coustom styles will create the styles.css file, the link to it is there for future reference.

meet_balls_app/tests.py
SchSeba: I don't see the change in the PR did you push?

Danielsio: I forgot to amend the cahnges

Yuval-Vino (approved):

SchSeba (approved):

PR #67, opened 2023-05-06
Add all game events page

### LINKED ISSUE #45: Court tests issues
In "test_create_three_courts_without_id":
1. Need to change variable names from court_id1 to court1.
2. Fix the check: Instead of checking IDs 1,2,3, I should check IDs: First, First+1, First+2.

I'll get to it ASAP.

### LINKED ISSUE #66: Make game-events.html to view all available games.
We can see the navbar in base.html.
In the navbar there is "Games" button. When you press it, you'll be routed to `http://127.0.0.1:8000/game-events`.
Currently, there is no such path available. 
@amitCohen2 and I will be working on this new feature.

ZvikaNaorCohen:
### Desired Outcome

Add all game events page. 

### Implemented Changes

- Add routing for `/game-events/`.
- Write function that displays all `GameEvents`.
- Add template for `game-events.html`.
- Fix court tests.
- Add filtering by min/max number of players.
- Add sort by every table entry.
- Add favicon.ico icon.

### Connected Issue/Story

closes #45 
closes #66 

### Dependencies

- None

### Definition of Done
- All `GameEvents` will be displayed, with an option for filtering, and when you press a specific ID, you'll be directed to another page.

#### Test coverage

- This PR is on Draft because there are no tests yet.

#### Documentation

- [ ] This PR does not require updating any documentation


ZvikaNaorCohen:
This is a draft PR. Feel free to comment it, but changes will be made in the next few days.
Thanks.

game_event/templates/game-events.html
```diff
@@ -0,0 +1,50 @@
...
+
+		th {
+			background-color: #f2f2f2;
+		}
+
+		tr:hover {
+			background-color: #f5f5f5;
+		}
+	</style>
+    <table>
```
amitCohen2, line 24: replace this: 
` <table>`
To this:
>     <script src="https://www.kryogenix.org/code/browser/sorttable/sorttable.js"></script>
>     <table class="sortable">
To make it sortable:)

Danielsio:
Excellent Work Naor and Amit, Code looks good to me, but I have a concern about the court.id, do we really want to display the id itself? How the users will even understand what it means ?
In my opinion "location" column should be displayed and the value can be the court.city or even a concat of the city and neighbourhood.
Don't you agree?

amitKaplansky:
What happens after a the time of a game event has passed? does the event stays in the table?

Danielsio:
> What happens after a the time of a game event has passed? does the event stays in the table?
> 
> 

I believe only "live events" must be displayed at least for the upcoming demo, so in the first place when querying for game events I would get all events that has time which isn't in the past and only display those events

I believe you can achieve this by using GameEvent.objects.filter(time > timezone.now())

But not sure syntax is correct.

ZvikaNaorCohen:
Thank you guys for your comments.
I added a checkbox to let the user choose wether he wants to see all available events, or hide the past events. Will be displayed in my upcoming commit.

Danielsio (approved):

ZvikaNaorCohen:
@Yuval-Vino @yftacherzog @OhadSeg @amitKaplansky @MaayanMashhadi @IdanZimi 
What do you guys think?

MaayanMashhadi:
> PR is ready to be reviewed. Tests have been added. The site at `127.0.0.1:8000/game-events` looks like this: WhatsApp Image 2023-05-10 at 11 36 49
> 
> There is an option to sort the table (click on the column name), filter by max,min players, max level, and check if you want to see past events or not. After this PR will be merged, we will be working on the UI to make the page more pretty.

Just for understanding: if I click on the id of the event, it will redirect to
the page of Amit and I responsible? (the information about the event)

ZvikaNaorCohen:
> Just for understanding about redirect to the information page: your purpose is if I click on the id of the event, it will redirect to the page of Amit and I responsible? (the information about the event)
Yes. And if we will have enough time, we will maybe add the same for the court: if we press the ID, we will be redirected to `127.0.0.1/court/{id}`.

game_event/views.py
```diff
@@ -1 +1,37 @@
...
+            max_players = int(request.GET.get('max_players'))
+            if max_players > 0:
+                game_events = game_events.filter(max_number_of_players__lte=max_players)
+        except ValueError:
+            pass
+
+    if request.GET.get('max_level'):
+        try:
+            max_level = int(request.GET.get('max_level'))
+            if max_level > 0:
```
MaayanMashhadi, line 28: I think the max_level should be 0 (included) to 11 (not included),
like the fields on class Rating.

MaayanMashhadi:
Hi Naor and Amit,
It looks great!
I left you some review and question and would like to hear you about it

MaayanMashhadi:
> > Just for understanding about redirect to the information page: your purpose is if I click on the id of the event, it will redirect to  the page of Amit and I responsible? (the information about the event)
> 
> Yes. And if we will have enough time, we will maybe add the same for the court: if we press the ID, we will be redirected to `127.0.0.1/court/{id}`.

Fantastic!

amitKaplansky:
Looks good for now, I think it look better if you put the filtering option in left and on right the table.

I had a different idea in my mind for the filtering,  adding a filter as a "Drop Down", I'll add a an example below, lmk what  you think.

ZvikaNaorCohen:
> Looks good for now, I think it look better if you put the filtering option in left and on right the table.
> 
> I had a different idea in my mind for the filtering, adding a filter as a "Drop Down", I'll add a an example below, lmk what you think.

I forgot to mention, we can't see it in the picture, but the min/max filter has increment/decrement arrows that are visible when we hover above the input box.
Also we can't see the sorting option. If we press the column name then an arrow will be shown to indicate if we want to sort in ascending or descending order.

game_event/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+from . import views
+
+
+urlpatterns = [
+    path('game-event/', views.game_events, name='game-events')
+    ]
```
Danielsio, line 7: Do you still need this? 

i dont know which PR will be merged first, so in a case this one get's merged, i will have to do quick rebase to include the game_event.urls in the meat_balls_urls

so i wonder maybe you can just fix it now  ? you can look at my PR #69 

and instead of the 'game-event' routing you got there you could replace that line with 
`path('', views.game_events, name='game events')`

the empty string in the path is because it has the prefix of "game-events/" from the meat_balls.urls

amitKaplansky:
In my point of view,

amitKaplansky:
In my point of view a thought that all the main pages should be located in the meet_balls_app directory, and all the pages  other pages should be added in a specific  model directory, for example game event page will be located in game_event directory with the tests and views that are related to that page.
what do you think?
@ZvikaNaorCohen @Danielsio @MaayanMashhadi @amitCohen2 @OhadSeg @IdanZimi @Yuval-Vino @yftacherzog

amitCohen2:
> In my point of view a thought that all the main pages should be located in the meet_balls_app directory, and all the pages other pages should be added in a specific model directory, for example game event page will be located in game_event directory with the tests and views that are related to that page. what do you think? @ZvikaNaorCohen @Danielsio @MaayanMashhadi @amitCohen2 @OhadSeg @IdanZimi @Yuval-Vino @yftacherzog

I think its a good idea to make it more organized

court/tests.py
```diff
@@ -28,15 +28,16 @@ def test_create_valid_court(self):
...
+        court1 = Court.objects.create(x=Decimal('11'), y=Decimal('22'),
+                                      city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
+        court2 = Court.objects.create(x=Decimal('22'), y=Decimal('33'),
+                                      city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=40)
+        court3 = Court.objects.create(x=Decimal('33'), y=Decimal('44'),
+                                      city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=50)
+        first_id = court1.courtID
+        assert court1.courtID == first_id
+        assert court2.courtID == first_id+1
+        assert court3.courtID == first_id+2
```
yftacherzog, line 40: You find yourself having to change this test in order to verify that court ID is as expected, but I don't think you actually have any obligations or trying to control the logic according to which IDs are being assigned.

If we're not trying to control the the IDs then what are we trying to test here? Are we trying to test that we can create courts without ID? Are trying to verify that they are getting new IDs? That there are actually new courts?

I suggest that you establish what is it that you're trying to test and then test that rather than testing the logic for assigning court IDs which I don't think we control nor we rely on. Suppose tomorrow you upgrade django and the new version defaults to assigning IDs randomly instead of with running numbers. You'll have regressions on this test that you had no reason to have.

game_event/views.py
```diff
@@ -1 +1,37 @@
...
+
+def game_events(request):
+    game_events = GameEvent.objects.all()
+
+    if request.GET.get('min_players'):
+        try:
+            min_players = int(request.GET.get('min_players'))
+            if min_players > 0:
+                game_events = game_events.filter(min_number_of_players__gte=min_players)
+        except ValueError:
```
yftacherzog, line 14: which line is expected to generate this exception? You should have only that line within the try block. Same goes for the filters below.

game_event/views.py
```diff
@@ -1 +1,37 @@
...
+from datetime import date
+
+
+def game_events(request):
+    game_events = GameEvent.objects.all()
+
+    if request.GET.get('min_players'):
+        try:
+            min_players = int(request.GET.get('min_players'))
+            if min_players > 0:
```
yftacherzog, line 12: Suppose that you drop this is statement, would it break anything? I suppose it will just make the next line run even if someone provides a non-positive number. If you drop it, I don't think there's a risk here of performance impact, but in return you get cleaner code (same goes for the other checks).

yftacherzog: Also, have in mind that every if-statement is another test case for your unit tests. Every time you generalize your code and remove flow of control statements or just branches of flow of control statements (e.g. 2 options instead of 3), you reduce the number of test cases you need and increase your code's maintainability.

game_event/test_game_event_views.py
```diff
@@ -0,0 +1,66 @@
...
+            time=timezone.now() - timezone.timedelta(days=1),
+            level_of_game=5,
+            min_number_of_players=4,
+            max_number_of_players=10,
+            court=saved_court,
+            ball_game='Basketball'
+        )
+        return game_event
+
+    def test_success_loading_site(self, client):
```
yftacherzog, line 30: You have a bunch of other tests that successfully load the page so I think what you need to test here is that you can load the page with no filters.
```suggestion
    def test_loading_site_no_filters(self, client):
```

game_event/test_game_event_views.py
```diff
@@ -0,0 +1,66 @@
...
+        response = client.get('/game-events/')
+        assert response.status_code == 200
+
+    def test_fail_loading_site(self, client):
+        response = client.get('/game-eventsss/')
+        assert response.status_code == 404
+
+    def test_valid_min_player_filter(self, client, saved_game_event):
```
yftacherzog, line 38: I think this test and the next ones need to be renamed. The way I see it, those tests are not about valid and invalid inputs. They are about filters that will or will not include your event. The tests are an important tool for clarifying the code, so if you call something invalid, it makes me think what makes this value invalid. In this case, I think those values are valid - they just don't return our event.

court/tests.py
ZvikaNaorCohen: This test was in the last demo, but with different variable names.
The problem that I had encountered is that I couldn't figure how to set the ID of the court, without assigning one.
I solved it by writing the function `def save(self, *args, **kwargs)` in `court/models.py`, which generates an ID for a court if none is assigned.
So in this test, what I'm testing is that if no ID is assigned, we'll get a generated one. So if 3 courts are made one after the other, they should have ascending order of IDs (firstID, firstID+1, firstID+2).

I agree that in some edge cases the test would fail (what if a new court is created between court1 is made, and before court2 is made? court2 will get firstID+2 instead of firstID+1). But if I remember correctly, in the previous PR (create court model) we agreed that this is a better implementation for the test, rather the older one.

game_event/test_game_event_views.py
ZvikaNaorCohen: Thanks for the reference. I was thinking about it while writing the code, something felt wrong for me, but I couldn't put my finger on it. Now I got it. Thanks.

game_event/urls.py
ZvikaNaorCohen: I'll change today according to your PR. I forgot to do it after we talked about it.

court/tests.py
yftacherzog: Alright. I didn't have that logic in mind. In that case, the test makes sense.

game_event/views.py
ZvikaNaorCohen: While writing this code I've got confused working parallelly on different files. In the HTML file we are limited, and can't get negative values for min/max inputs. It's not shown on the screenshots I've uploaded, but when we hover above the min/max inputs, there are arrows for plus/minus. So the user will be using these arrows to give us the input, and we won't get any "wrong" values that will cause our program to crash.

After deleting the "try,except,if" statements, the code looks much better and functions the same.
Thank you for the reference.

ZvikaNaorCohen:
Hello! I've fixed this PR according to your comments. Does anyone have any last comment, or can we merge it? I would like to work on the CSS file right after this PR will be merged.

Yuval-Vino (approved):

MaayanMashhadi (approved):

game_event/views.py
yftacherzog: @MaayanMashhadi can this one be resolved?

game_event/urls.py
yftacherzog: @Danielsio can this one be resolved?

ZvikaNaorCohen:
@yftacherzog Rebased and solved the conflicts.
I'm working on solving the two unresolved conversations. For some reason I can't even see them. I'll figure it out now.

game_event/views.py
MaayanMashhadi: > @MaayanMashhadi can this one be resolved?

Yes, but unfortunately I didn't success to resolve this conversation.

PR #69, opened 2023-05-07
Add create_game_event template

### LINKED ISSUE #68: Add template for a Form to create game event
## Expected Behavior

users will have template that contains form fields to create game events 

## Current Behavior

no such template exist

## Acceptance Criteria

1. create_game_event template must be added
2. routing for template must be added

Danielsio:
### Desired Outcome

1. users will have a template to create game events
2. the template will have routing like "/game-events/create"

### Implemented Changes
1. I added a template for the form to create game events.
2. added a view that will render the template + will accept the post reqeust from the user and create the game event
3. added basic validation  for the view
4. applied group routing for game-events 

Most of the PR is the game_events.views.create_game_event function
and the template itself game_event\templates\game_event\create_game_event.html

### Connected Issue/Story

Resolves #68 

### Dependencies

There are no dependencies for now.

### Definition of Done

1. Form for creation of GameEvents for the Players will be displayed at routing **`"game-events/create"`**
2. A view will be created which handles the display of the form and the creation GameEvent from the filled form by the Player.
3. Form Data will be validated inside the **`GameEvent.create()`** function, and if valid will create the GameEvent. 
4. Otherwise error messeages will be displayed and the player will be **`redirected back to the form page`**.

#### Test coverage

- [x] This PR includes new unit and integration tests to go with the code changes

#### Documentation

### DIFF
--- a/court_ball_game/models.py
+++ b/court_ball_game/models.py
@@ -12,3 +12,8 @@ class CourtBallGame(models.Model):
 
     class Meta:
         unique_together = ('court', 'ball_game')
+
+    @staticmethod
+    def is_ball_game_playable(court, ball_game):
+        ball_games_available_at_court = CourtBallGame.objects.filter(court=court)
+        return any(bg.ball_game == ball_game for bg in ball_games_available_at_court)
--- a/court_ball_game/tests.py
+++ b/court_ball_game/tests.py
@@ -59,3 +59,10 @@ class TestCourtBallGameModel:
             CourtBallGame.objects.create(
                 court=court_instance
             ).full_clean()
+
+    def test_is_ball_game_playable(self, court_instance, court_ball_game_instance):
+        assert CourtBallGame.is_ball_game_playable(court_instance, BallGame.Basketball)
+
+    def test_is_ball_game_playable_false(self, court_instance):
+        expected = False
+        assert expected == CourtBallGame.is_ball_game_playable(court_instance, BallGame.Basketball)
--- a/game_event/models.py
+++ b/game_event/models.py
@@ -3,6 +3,10 @@ from django.core.validators import MinLengthValidator
 from court.models import Court
 from player_rating.models import Rating
 from player.models import BallGame
+from court_ball_game.models import CourtBallGame
+from django.utils import timezone
+from django.core.exceptions import ValidationError
+import datetime
 
 
 class GameEvent(models.Model):
@@ -17,3 +21,33 @@ class GameEvent(models.Model):
         validators=[MinLengthValidator(4)],
         choices=BallGame.choices
     )
+
+    @staticmethod
+    def create(time, level_of_game, min_number_of_players, max_number_of_players, court, ball_game):
+        GameEvent.validate_game_event(court, ball_game, min_number_of_players, max_number_of_players, time)
+        game_event = GameEvent(time=time,
+                               level_of_game=level_of_game,
+                               min_number_of_players=min_number_of_players,
+                               max_number_of_players=max_number_of_players,
+                               court=court,
+                               ball_game=ball_game)
+        game_event.save()
+        return game_event
+
+    @staticmethod
+    def validate_game_event(court, ball_game, min_number_of_players, max_number_of_players, time):
+        errors = []
+        if not GameEvent.is_ball_game_playable_at_court(court, ball_game):
+            errors.append("The selected court doesn't support the selected ball game.")
+        if min_number_of_players < 2:
+            errors.append("The minimum number of players must be at least 2.")
+        if max_number_of_players < min_number_of_players:
+            errors.append("The maximum number of players must be at least equal to the minimum number of players")
+        if time < timezone.now() or time > timezone.now() + datetime.timedelta(days=30):
+            errors.append("The selected time must be between now and one month from now.")
+        if errors:
+            raise ValidationError("\n".join(errors))
+
+    @staticmethod
+    def is_ball_game_playable_at_court(court, ball_game):
+        return CourtBallGame.is_ball_game_playable(court, ball_game)
--- a/game_event/tests.py
+++ b/game_event/tests.py
@@ -4,6 +4,10 @@ from court.models import Court
 from decimal import Decimal
 from django.utils import timezone
 from player.models import BallGame
+from django.urls import reverse
+from .views import process_game_event_form
+from court_ball_game.models import CourtBallGame
+from datetime import timedelta
 import pytest
 
 TEST_ID = 2
@@ -28,6 +32,10 @@ class TestGameEventModel:
         return GameEvent.objects.create(id=TEST_ID, time=TEST_TIME, level_of_game=TEST_LEVEL,
                                         min_number_of_players=TEST_MIN, max_number_of_players=TEST_MAX,
                                         court=saved_court, ball_game=TEST_BALL_GAME)
+
+    @pytest.fixture
+    def court_ball_game(self, saved_court):
+        return CourtBallGame.objects.create(court=saved_court, ball_game=TEST_BALL_GAME)
 
     def test_get_game_event(self, saved_game_event):
         game_event = GameEvent.objects.get(pk=saved_game_event.pk)
@@ -75,3 +83,99 @@ class TestGameEventModel:
             GameEvent.objects.create(id=TEST_ID, time=TEST_TIME, level_of_game=TEST_LEVEL,
                                      min_number_of_players=TEST_MIN, max_number_of_players=TEST_MAX,
                                      court=saved_court,).full_clean()
+
+    def test_create_game_event_valid_form(self, client, saved_court, court_ball_game):
+        count_of_game_events_before = GameEvent.objects.count()
+        url = reverse(process_game_event_form)
+        time_now_plus_one_day = (timezone.now() + timedelta(days=1)).strftime('%Y-%m-%dT%H:%M')
+        form_data = {
+            'time': time_now_plus_one_day,
+            'level_of_game': 5,
+            'min_number_of_players': 2,
+            'max_number_of_players': 4,
+            'court': saved_court.courtID,
+            'ball_game': TEST_BALL_GAME,
+        }
+        client.post(url, data=form_data)
+        assert GameEvent.objects.count() == count_of_game_events_before + 1
+
+    def test_validate_game_event_creation_bad_time(self, saved_court, court_ball_game):
+        bad_test_time = timezone.datetime.fromisoformat("2022-05-10T14:30")
+        bad_test_time = timezone.make_aware(bad_test_time)
+        with pytest.raises(ValidationError):
+            GameEvent.validate_game_event(
+                court=saved_court.courtID,
+                ball_game=TEST_BALL_GAME,
+                min_number_of_players=2,
+                max_number_of_players=4,
+                time=bad_test_time)
+
+    def test_validate_game_event_creation_bad_ball_game(self, saved_court, court_ball_game):
+        time_now_plus_one_day = timezone.now() + timedelta(days=1)
+        with pytest.raises(ValidationError):
+            GameEvent.validate_game_event(
+                court=saved_court.courtID,
+                ball_game="Golf",
+                min_number_of_players=2,
+                max_number_of_players=4,
+                time=time_now_plus_one_day)
+
+    def test_validate_game_event_creation_negative_min_players(self, saved_court, court_ball_game):
+        time_now_plus_one_day = timezone.now() + timedelta(days=1)
+        with pytest.raises(ValidationError):
+            GameEvent.validate_game_event(
+                court=saved_court.courtID,
+                ball_game=TEST_BALL_GAME,
+                min_number_of_players=-1,
+                max_number_of_players=4,
+                time=time_now_plus_one_day)
+
+    def test_validate_game_event_creation_max_players_less_then_min_players(self, saved_court, court_ball_game):
+        time_now_plus_one_day = timezone.now() + timedelta(days=1)
+        with pytest.raises(ValidationError):
+            GameEvent.validate_game_event(
+                court=saved_court.courtID,
+                ball_game=TEST_BALL_GAME,
+                min_number_of_players=2,
+                max_number_of_players=1,
+                time=time_now_plus_one_day)
+
+    def test_validate_game_event_creation(self, saved_court, court_ball_game):
+        time_now_plus_one_day = timezone.now() + timedelta(days=1)
+        GameEvent.validate_game_event(
+            court=saved_court.courtID,
+            ball_game=TEST_BALL_GAME,
+            min_number_of_players=2,
+            max_number_of_players=10,
+            time=time_now_plus_one_day)
+
+    def test_game_event_create(self, saved_court, court_ball_game):
+        count_of_game_events = GameEvent.objects.count()
+        time_now_plus_one_day = timezone.now() + timedelta(days=1)
+        level_of_game = 5
+        min_number_of_players = 2
+        max_number_of_players = 5
+        court = saved_court
+        ball_game = TEST_BALL_GAME
+        GameEvent.create(
+            time_now_plus_one_day,
+            level_of_game,
+            min_number_of_players,
+            max_number_of_players,
+            court, ball_game)
+        assert GameEvent.objects.count() == count_of_game_events + 1
+
+    def test_post_game_event(self, client, saved_court, court_ball_game):
+        count_of_game_events = GameEvent.objects.count()
+        time_now_plus_one_day = (timezone.now() + timedelta(days=1)).strftime('%Y-%m-%dT%H:%M')
+        response = client.post('/game-events/create/process', {
+            'time': time_now_plus_one_day,
+            'level_of_game': '5',
+            'min_number_of_players': '3',
+            'max_number_of_players': '10',
+            'court': saved_court.courtID,
+            'ball_game': TEST_BALL_GAME,
+        })
+        assert GameEvent.objects.count() == count_of_game_events + 1
+        assert response.status_code == 302
+        assert response.url == '/'
--- a/game_event/urls.py
+++ b/game_event/urls.py
@@ -1,0 +1,8 @@
+from django.urls import path
+from . import views
+
+
+urlpatterns = [
+    path('create/', views.display_game_event_form, name='display game event form'),
+    path('create/process', views.process_game_event_form, name='process game event form'),
+    ]
--- a/game_event/views.py
+++ b/game_event/views.py
@@ -1,1 +1,42 @@
-# from django.shortcuts import render
+from .models import GameEvent
+from django.shortcuts import render, redirect
+from court.models import Court
+from player_rating.models import Rating
+from player.models import BallGame
+from django.utils import timezone
+from django.core.exceptions import ValidationError
+from django.contrib import messages
+
+
+def display_game_event_form(request):
+    courts = Court.objects.all()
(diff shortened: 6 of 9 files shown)


meetballs/urls.py
```diff
@@ -14,11 +14,12 @@
...
-from meet_balls_app import views
+from django.urls import path, include
+from meet_balls_app import views as app_views
 
 
 urlpatterns = [
     path('admin/', admin.site.urls),
-    path('', views.home, name='home'),
+    path('', app_views.home, name='home'),
+    path('game-events/', include("game_event.urls"))
```
amitCohen2, line 24: May be better to add at the end of this line ','  to avoid problems later.
Should look like that this line:  path('game-events/', include("game_event.urls")),

Danielsio: Sure

game_event/views.py
```diff
@@ -1 +1,56 @@
...
+    else:
+        courts = Court.get_all_courts()
+        return render(request, 'game_event/create_game_event.html', {
+            'courts': courts,
+            'ratings': Rating.choices,
+            'games': BallGame.choices
+        })
+
+
+def validate_game_event_form(request, court, ball_game, min_number_of_players, time):
```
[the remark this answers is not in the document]
Danielsio: i am not sure how to test this to be honest.. @yftacherzog @Yuval-Vino maybe you know simple way to verify that form logic is ok ?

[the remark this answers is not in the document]
Yuval-Vino: In my opinion, you can move all validators to the create function of the class (which is missing btw). Then, in the `create_game_event `function, you could try to create the game event and catch any errors that occur.

After that, you could add some tests to the create function to ensure that it works correctly.

For the views tests, you could send a valid request and check that the game event is created, and also send an invalid request and make sure that the game event hasn't been created.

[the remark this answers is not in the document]
Danielsio: I managed to get it to work somehow

game_event/models.py
```diff
@@ -17,3 +20,45 @@ class GameEvent(models.Model):
         validators=[MinLengthValidator(4)],
         choices=BallGame.choices
     )
+
+    @staticmethod
+    def create_game_event_from_form(request):
```
Yuval-Vino, line 25: My intention was to create a function (similar to the one in #47) that incorporates validators.
This way, from the views, you can simply call the create method(which shouldnt require a request input, but only relevant fields for `GameEvent`). If an error occurs during the creation process, it can be caught in the views and a relevant error message can be sent

This way the creation of `GameEvent` will be more robust, and not force it to be in a request form

game_event/tests.py
```diff
@@ -75,3 +82,32 @@ def test_create_game_event_with_null_ball_game(self, saved_court):
...
+        form_data = {
+            'time': "2034-05-10T14:30",
+            'level_of_game': 5,
+            'min_number_of_players': 2,
+            'max_number_of_players': 4,
+            'court': saved_court.courtID,
+            'ball_game': TEST_BALL_GAME,
+        }
+        client.post(url, data=form_data)
+        assert GameEvent.objects.count() == 1
```
Yuval-Vino, line 97: This test will fail when you have data. To prevent this, consider adding the previous size of the data as a variable before running the test. Then, you can test that the new size is equal to the previous size plus one

game_event/tests.py
```diff
@@ -75,3 +82,32 @@ def test_create_game_event_with_null_ball_game(self, saved_court):
...
+        data = {
+            'time': "2022-05-10T14:30",
+            'level_of_game': 5,
+            'min_number_of_players': -1,
+            'max_number_of_players': 4,
+            'court': saved_court.courtID,
+            'ball_game': TEST_BALL_GAME,
+        }
+        client.post(url, data=data)
+        assert GameEvent.objects.count() == 0
```
Yuval-Vino, line 110: same

court_ball_game/models.py
```diff
@@ -12,3 +12,11 @@ class CourtBallGame(models.Model):
...
     class Meta:
         unique_together = ('court', 'ball_game')
+
+    @staticmethod
+    def is_ball_game_playable(court, ball_game):
+        ball_games_available_at_court = CourtBallGame.objects.filter(court=court)
+        for bg in ball_games_available_at_court:
+            if bg.ball_game == ball_game:
+                return True
+        return False
```
Yuval-Vino, line 22: More pythonic
```suggestion
    return any(bg.ball_game == ball_game for bg in ball_games_available_at_court)

```

Yuval-Vino (changes requested):

game_event/tests.py
Danielsio: Fixed

game_event/tests.py
Danielsio: Fixed

court_ball_game/models.py
Danielsio: Thanks

game_event/models.py
Danielsio: i fixed it , but in this current approch only one messeage will be displayed to the user.. the first ValidationError.

game_event/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+from . import views
+
+
+urlpatterns = [
+    path('create/', views.create_game_event, name='create game event'),
```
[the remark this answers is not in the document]
Danielsio: Thanks, That is a really good advice, will be changed to display_game_event_form.

game_event/models.py
Yuval-Vino: I suggest modifying the function to throw an exception with a string instead of a list. This will make it more user-friendly for future users who may expect a string output.
To achieve this, you can init errors to an empty string and concatenating relevant errors using `errors += 'relevant error'` instead of repeatedly appending to the list. 
Alternatively, if you want to keep the current Implementation you can use the join method to concatenate the errors list  
  `" ".join(errors)`.

What do you think?

game_event/models.py
```diff
@@ -17,3 +20,27 @@ class GameEvent(models.Model):
...
+                               level_of_game=level_of_game,
+                               min_number_of_players=min_number_of_players,
+                               max_number_of_players=max_number_of_players,
+                               court=court,
+                               ball_game=ball_game)
+        game_event.save()
+        return game_event
+
+    @staticmethod
+    def validate_game_event(court, ball_game, min_number_of_players, time):
```
Yuval-Vino, line 37: why dont we validate `max_number_of_players`?

game_event/models.py
```diff
@@ -17,3 +20,27 @@ class GameEvent(models.Model):
...
+        errors = []
+        if not CourtBallGame.is_ball_game_playable(court, ball_game):
+            errors.append("The selected court doesn't support the selected ball game.")
+        if min_number_of_players < 1:
+            errors.append("The minimum number of players must be at least 1.")
+        if time < timezone.now():
+            errors.append("The selected time must be in the future.")
```
MaayanMashhadi, line 44: Maybe consider to add upper bound about the time of the event.
I think maybe to add event just in the current month. (Cause I think to start some event in the far future, like more than half a year isn't so good, it could mess up the web). 
Would like to hear your opinion about it.

MaayanMashhadi:
Hi Daniel,
You did a good job here!
I left you some reviews, would like to hear what you think about it.

game_event/models.py
Danielsio: could you elaborate ? I didnt understand the difference in your approch.

game_event/models.py
Danielsio: you are right, i can validate if max - min is > 0

game_event/models.py
```diff
@@ -17,3 +20,27 @@ class GameEvent(models.Model):
...
+                               ball_game=ball_game)
+        game_event.save()
+        return game_event
+
+    @staticmethod
+    def validate_game_event(court, ball_game, min_number_of_players, time):
+        errors = []
+        if not CourtBallGame.is_ball_game_playable(court, ball_game):
+            errors.append("The selected court doesn't support the selected ball game.")
+        if min_number_of_players < 1:
```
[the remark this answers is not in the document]
Danielsio: sounds good

game_event/models.py
Danielsio: i will change it to a month interval

MaayanMashhadi: > i will change it to a month interval

I'm not sure you understand me correct so just for ensuring.
I mean to add another validate about the future (not the past), like something like this:
`if time.month > timezone.now().month+1: error`
this validate that the event will be in the current month or month after and not be held about 6 month after...

game_event/models.py
Danielsio: It's a little uncomfortable this way, cause i want to later iterate over the errors and print messeage to the user

Danielsio:
image

here are some screenshots of the form and the possible validation erros.

image

And here is just noraml form

Danielsio:
Dear Reviewers. 
It's Important that you let me know, if the current implementation of the value of the court is sufficient (for future perposes maybe), since it's only the concatenation of the city and the neighborhood, and I thought it would be more user friendly fot the creator of an event to understand what court he is choosing .

ZvikaNaorCohen:
In my opinion, we should keep the court value like that.

game_event/tests.py
```diff
@@ -75,3 +84,32 @@ def test_create_game_event_with_null_ball_game(self, saved_court):
...
+
+    def test_create_game_event_invalid_form(self, client, saved_court):
+        count_of_game_events_before = GameEvent.objects.count()
+        url = reverse(display_game_event_form)
+        data = {
+            'time': "2022-05-10T14:30",
+            'level_of_game': 5,
+            'min_number_of_players': -1,
+            'max_number_of_players': 4,
+            'court': saved_court.courtID,
```
ZvikaNaorCohen, line 111: I think that `GameEvent` entity has a data member of court (`court = models.ForeignKey(Court, on_delete=models.CASCADE)`) and not `court.courtID`. 
It raised a few errors to me when I was dealing with creating new GameEvent entity. 
If we remove the .courtID and keep just the saved_court, will the tests fail?

ZvikaNaorCohen:
> image
> 
> here are some screenshots of the form and the possible validation erros.
> 
> image
> 
> And here is just noraml form

The validation tests look very good with that red warning. Well done.
What happens when a user wants to enter a court? Does he have a list of all the existing ones, and he picks from there?
Or does he write his own text, and a court would be created?

game_event/models.py
```diff
@@ -17,3 +21,29 @@ class GameEvent(models.Model):
...
+
+    @staticmethod
+    def validate_game_event(court, ball_game, min_number_of_players, max_number_of_players, time):
+        errors = []
+        if not CourtBallGame.is_ball_game_playable(court, ball_game):
+            errors.append("The selected court doesn't support the selected ball game.")
+        if min_number_of_players < 2:
+            errors.append("The minimum number of players must be at least 2.")
+        if max_number_of_players - min_number_of_players < 0:
+            errors.append("The Maximum number of players must be at least equal to the minimum number of players")
```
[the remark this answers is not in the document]
Danielsio: thanks, nice catch

Danielsio:
> The validation tests look very good with that red warning. Well done. What happens when a user wants to enter a court? Does he have a list of all the existing ones, and he picks from there? Or does he write his own text, and a court would be created?

For now, i render select option for each instance of a scourt, so user chooses from the existing courts in db.
maybe in the future we could manage a field for the user to input some address e.g "Kriat Ono Shlomo hameleh" and with the help of th "pin" library we can extract longitude and latitude from it, to create fresh instance of court in db. just a thought

game_event/tests.py
Danielsio: i belive you are right

Danielsio: i get Field 'courtID' expected a number but got 'Court object (1)'. when i tried to send him only the saved_court

MaayanMashhadi:
> Dear Reviewers. It's Important that you let me know, if the current implementation of the value of the court is sufficient (for future perposes maybe), since it's only the concatenation of the city and the neighborhood, and I thought it would be more user friendly fot the creator of an event to understand what court he is choosing .

For me it sounds good, I agree it will be more user friendly.

Danielsio:
Do you still need your routing at game_event.urls ? 

i dont know which PR will be merged first, so in a case this one get's merged, i will have to do quick rebase to include the game_event.urls in the meat_balls_urls

so i wonder maybe you can just fix it now  ? you can look at my PR #69 

and instead of the 'game-event' routing you got there you could replace that line with 
`path('', views.game_events, name='game events')`

the empty string in the path is because it has the prefix of "game-events/" from the meat_balls.urls

court_ball_game/models.py
```diff
@@ -12,3 +12,8 @@ class CourtBallGame(models.Model):
 
     class Meta:
         unique_together = ('court', 'ball_game')
+
+    @staticmethod
+    def is_ball_game_playable(court, ball_game):
```
yftacherzog, line 17: missing tests for this one

court/models.py
```diff
@@ -15,3 +15,7 @@ def save(self, *args, **kwargs):
             max_court_id = Court.objects.aggregate(models.Max('courtID'))['courtID__max'] or 0
             self.courtID = max_court_id + 1
         super().save(*args, **kwargs)
+
+    @staticmethod
+    def get_all_courts():
```
yftacherzog, line 20: missing tests for this one

game_event/models.py
```diff
@@ -17,3 +21,29 @@ class GameEvent(models.Model):
...
+                               ball_game=ball_game)
+        game_event.save()
+        return game_event
+
+    @staticmethod
+    def validate_game_event(court, ball_game, min_number_of_players, max_number_of_players, time):
+        errors = []
+        if not CourtBallGame.is_ball_game_playable(court, ball_game):
+            errors.append("The selected court doesn't support the selected ball game.")
+        if min_number_of_players < 2:
```
yftacherzog, line 42: Can't this validation be part of the model?

game_event/models.py
```diff
@@ -17,3 +21,29 @@ class GameEvent(models.Model):
...
+                               level_of_game=level_of_game,
+                               min_number_of_players=min_number_of_players,
+                               max_number_of_players=max_number_of_players,
+                               court=court,
+                               ball_game=ball_game)
+        game_event.save()
+        return game_event
+
+    @staticmethod
+    def validate_game_event(court, ball_game, min_number_of_players, max_number_of_players, time):
```
yftacherzog, line 38: missing tests

game_event/models.py
```diff
@@ -17,3 +21,29 @@ class GameEvent(models.Model):
         validators=[MinLengthValidator(4)],
         choices=BallGame.choices
     )
+
+    @staticmethod
+    def create(time, level_of_game, min_number_of_players, max_number_of_players, court, ball_game):
```
yftacherzog, line 26: missing tests

game_event/views.py
```diff
@@ -1 +1,38 @@
...
+from django.shortcuts import render, redirect
+from court.models import Court
+from player_rating.models import Rating
+from player.models import BallGame
+from django.utils import timezone
+from django.core.exceptions import ValidationError
+from django.contrib import messages
+
+
+def display_game_event_form(request):
```
yftacherzog, line 11: missing tests.

yftacherzog: Also, by having the same function used for both http methods, you're making this code harder to test. i.e. harder to maintain.

game_event/tests.py
```diff
@@ -75,3 +84,32 @@ def test_create_game_event_with_null_ball_game(self, saved_court):
...
+            'level_of_game': 5,
+            'min_number_of_players': 2,
+            'max_number_of_players': 4,
+            'court': saved_court.courtID,
+            'ball_game': TEST_BALL_GAME,
+        }
+        client.post(url, data=form_data)
+        assert GameEvent.objects.count() == count_of_game_events_before + 1
+
+    def test_create_game_event_invalid_form(self, client, saved_court):
```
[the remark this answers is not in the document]
Danielsio: I replaced this test with several validation tests

game_event/models.py
Danielsio: Added tests

game_event/tests.py
```diff
@@ -4,6 +4,11 @@
 from decimal import Decimal
 from django.utils import timezone
 from player.models import BallGame
+from django.urls import reverse
+from .views import display_game_event_form  # import the view function
```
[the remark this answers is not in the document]
Danielsio: Nope

game_event/models.py
Danielsio: not sure i understood
i wrap the call with the GameEvent.is_ball_game_playable_at_court()
if it wasnt your intension then i would need elaboration on this part

court/models.py
Danielsio: added one

court_ball_game/models.py
Danielsio: added test for this

game_event/views.py
Danielsio: took me some modifications, but i managed to split it to seperate views

Danielsio: added test for display the form, and for post process

court/tests.py
```diff
@@ -127,3 +127,7 @@ def test_court_without_max_players(self):
                 city='TEST_CITY',
                 neighborhood='TEST_NEIGHBORHOOD'
             )
+
+    def test_get_all_courts(self, court):
+        courts = Court.get_all_courts()
+        assert court in courts
```
yftacherzog, line 133: You're testing that court is there rather than all courts. To test it in a robust way, you'd need to use the actual line that you use inside the method. That raises the question, do you really gain anything by adding this method? You could just replace any call to this custom method to a call to the built-in method that this custom method is calling. No? It will be just as easy to use and will mean less code to maintain.

court_ball_game/tests.py
```diff
@@ -59,3 +59,6 @@ def test_create_court_ball_game_with_null_ball_game(self, court_instance):
             CourtBallGame.objects.create(
                 court=court_instance
             ).full_clean()
+
+    def test_is_ball_game_playable(self, court_instance, court_ball_game_instance):
```
yftacherzog, line 63: You proved here that it can return true. You need to also prove that it can return false (another test case or parametrize this test with pytest parametrize).

game_event/views.py
```diff
@@ -1 +1,42 @@
...
+            GameEvent.create(time, level_of_game, min_number_of_players, max_number_of_players, court, ball_game)
+        except ValidationError as e:
+            error_messages = e.messages[0].split("\n")
+            for error in error_messages:
+                messages.error(request, error)
+            return redirect("/game-events/create")
+
+        return redirect("/")  # will be changes later to "/game-events"
+    else:
+        redirect("/game-events/create")
```
yftacherzog, line 42: What's the reason here it's `redirect` and earlier it's `return redirect`?

court/tests.py
Danielsio: True, will be replaced

game_event/views.py
Danielsio: my mistake

court_ball_game/tests.py
Danielsio: added test with expected false

Danielsio:
@Yuval-Vino I hope I have addressed all your comments, you can update me if something remains unaddressed and I would be happy if you resolve the conversations.

Yuval-Vino (approved):

yftacherzog (approved):

PR #73, opened 2023-05-10
Fix player_rating test

### LINKED ISSUE #72: Fix player_rating test
### Expected Behavior
The test `test_delete_player_deletes_player_ratings` at `player_rating/tests.py` should pass also when we have a real data in out DB. 

### Current Behavior
If we have a real users/players in our DB the test fails because it uses hard coded numbers. It didn't take into consideration other users exist other than the fixture.

### Steps to Reproduce

- Fix the test

IdanZimi:
### Desired Outcome

Fixed the 'test_delete_player_deletes_player_ratings' at player_rating/tests.py. The test will pass now also when we have real data in DB.

### Implemented Changes

Changed the `test_delete_player_deletes_player_ratings` at `player_rating/tests.py`

### Connected Issue/Story

closes #72 

### Dependencies

This section inidicates whether this PR depends on any other PR's. 
Might need to wait for other PR's to merged before. Add the PR ID.

### Definition of Done
The test test_delete_player_deletes_player_ratings at player_rating/tests.py should pass also when we have a real data in out DB. Thus, changed the test.

#### Test coverage

  changes
- [ ] The changes in this PR do not require tests

#### Documentation

### DIFF
--- a/player_rating/tests.py
+++ b/player_rating/tests.py
@@ -1,5 +1,5 @@
 from .models import PlayerRating, Rating
-from player.models import BallGame, Player
+from player.models import BallGame
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError
 import pytest
@@ -66,11 +66,11 @@ class TestPlayerRatingModel:
                 rating=0)
 
     def test_delete_player_deletes_player_ratings(self, player, player_rating):
-        assert Player.objects.count() == 1
-        assert PlayerRating.objects.count() == 1
+        all_players_ratings_count_before = PlayerRating.objects.count()
+        player_fixture_ratings_count = PlayerRating.objects.filter(player=player).count()
+        assert player_fixture_ratings_count > 0
         player.delete()
-        assert Player.objects.count() == 0
-        assert PlayerRating.objects.count() == 0
+        assert PlayerRating.objects.count() == all_players_ratings_count_before - player_fixture_ratings_count
 
     def test_create_valid_player_rating(self, player):
         PlayerRating.objects.create(


Danielsio (approved):
Well Done ! the test is more much more durable now.

OhadSeg (approved):

player_rating/tests.py
```diff
@@ -66,11 +66,14 @@ def test_create_player_rating_invalid_rating2(self, player_rating):
...
-        assert PlayerRating.objects.count() == 1
+        all_players_ratings_before = PlayerRating.objects.count()
+        player_fixture_ratings = PlayerRating.objects.filter(player=player).count()
+        assert Player.objects.filter(user=player.user).exists()
+        assert PlayerRating.objects.count() > 0
         player.delete()
-        assert Player.objects.count() == 0
-        assert PlayerRating.objects.count() == 0
+        with pytest.raises(Player.DoesNotExist):
+            Player.objects.get(pk=player.pk)
```
Yuval-Vino, line 75: There's no need to test this as it has already been verified in `test_delete_player `in the player model tests

player_rating/tests.py
```diff
@@ -66,11 +66,14 @@ def test_create_player_rating_invalid_rating2(self, player_rating):
                 rating=0)
 
     def test_delete_player_deletes_player_ratings(self, player, player_rating):
-        assert Player.objects.count() == 1
-        assert PlayerRating.objects.count() == 1
+        all_players_ratings_before = PlayerRating.objects.count()
+        player_fixture_ratings = PlayerRating.objects.filter(player=player).count()
+        assert Player.objects.filter(user=player.user).exists()
+        assert PlayerRating.objects.count() > 0
```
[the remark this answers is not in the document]
IdanZimi: to check if the `player_rating` fixture was created. I see now there is a test for it at `test_get_player_rating`.
I'll delete it.

player_rating/tests.py
IdanZimi: removed

SchSeba (approved):

player_rating/tests.py
```diff
@@ -66,11 +66,11 @@ def test_create_player_rating_invalid_rating2(self, player_rating):
                 rating=0)
 
     def test_delete_player_deletes_player_ratings(self, player, player_rating):
-        assert Player.objects.count() == 1
-        assert PlayerRating.objects.count() == 1
+        all_players_ratings_before = PlayerRating.objects.count()
+        player_fixture_ratings = PlayerRating.objects.filter(player=player).count()
```
yftacherzog, line 70: What if this one returns zero? If we assert that `player_fixture_ratings` is greater than zero then we can address this issue and also make the line below redundant. i.e. you could remove this code `assert Player.objects.filter(user=player.user).exists()`.

IdanZimi: @yftacherzog how it can return zero? isn't it a fixture that created for the test? `test_get_player_rating` is checking if the fixture created. what am I missing?

player_rating/tests.py
```diff
@@ -66,11 +66,11 @@ def test_create_player_rating_invalid_rating2(self, player_rating):
                 rating=0)
 
     def test_delete_player_deletes_player_ratings(self, player, player_rating):
-        assert Player.objects.count() == 1
-        assert PlayerRating.objects.count() == 1
+        all_players_ratings_before = PlayerRating.objects.count()
```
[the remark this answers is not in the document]
IdanZimi: renamed

player_rating/tests.py
IdanZimi: I think the code `assert Player.objects.filter(user=player.user).exists()` is redundant even without asserting `player_fixture_ratings` is greater than zero no?

yftacherzog: > I think the code `assert Player.objects.filter(user=player.user).exists()` is redundant even without asserting `player_fixture_ratings` is greater than zero no?

Yes. I think it's redundant anyway.


> @yftacherzog how it can return zero? isn't it a fixture that created for the test? `test_get_player_rating` is checking if the fixture created. what am I missing?

The fixture is making it so that it won't be zero. Adding that line can make the test more self explanatory. By adding that single line you are making it more clear what you're testing there without having to go elsewhere to figure out what the fixture is about. 

Having said that, it's not a must to have it in this case. It's a matter of preference. I'll let you decide.

IdanZimi: Ok thanks. I changed the test to include suggested assertion.

Yuval-Vino (approved):

yftacherzog (approved):

PR #77, opened 2023-05-14
Add game event template

### LINKED ISSUE #76: Add game event template
### Expected Behavior

users will have template that contains a specific  game event.

### Current Behavior

no such template exist

Acceptance Criteria

game_event template must be added
routing for template must be added

amitKaplansky:
### Desired Outcome

Add game event page.

### Implemented Changes

1. Write function that displays `GameEvent.`
2. Add templates directory in game-event for GameEvent template:
Added in directory:
- game-event.html
- join-event.html
- result/html
4. Add tests for game-event template
5. added routing for `/game-event/` in meatballs/urls.py and in game-event/urls.py.

### Connected Issue/Story

closes #76 

### Dependencies

### Definition of Done

- `GameEvent` will be displayed, with option to join to the event or removing yourself from an event.

#### Test coverage

- This PR has tests related to the game-event template.

#### Documentation

- [ ] This PR does not require updating any documentation

### DIFF
--- a/game_event/templates/game_event/game-event.html
+++ b/game_event/templates/game_event/game-event.html
@@ -1,0 +1,73 @@
+{% extends 'base.html' %}
+{% block content %}
+{% load static %}
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Game Event</title>
+    <link rel="stylesheet" type="text/css" href="{% static 'css/game-event.css' %}">
+    <script>
+        setTimeout(function() {
+            location.reload();
+        }, 5000);
+    </script>
+  </head>
+  <body>
+    {% if time %}
+        <h1>Event Details</h1>
+        <p>Time: {{ time }}</p>
+        <p>Level: {{ level_of_game }}</p>
+        <p>Min Players: {{ min_number_of_players }}</p>
+        <p>Max Players: {{ max_number_of_players }}</p>
+        <p>Court: {{ court }} {{ neighborhood }}</p>
+        <p>Game: {{ ball_game }}</p>
+        <h2>Players:</h2>
+        <table class="nice-table">
+            <thead>
+                <tr>
+                <th> </th>
+                <th>Name</th>
+                <th>Brings Ball</th>
+            </tr>
+            </thead>
+                <tbody>
+                    {% for player in event_players %}
+                        <tr>
+                            <td>
+                                <img src="{% static 'images/favicon.ico' %}" alt="Ball Image" class="ball-image">
+                            </td>
+                            <td>{{ player.first_name }} {{ player.last_name }}</td>
+                            <td>
+                                {% if player.brings_ball == True %}
+                                    Yes
+                                {% else %}
+                                    No
+                                {% endif %}
+                            </td>
+                        </tr>
+                    {% endfor %}
+                </tbody>
+            </table>
+            <div>
+                {% if in_event %}
+                    <a href="/game-events/remove-from-event/{{ id }}/"
+                    class="round-button"
+                    onclick="window.open(this.href, 'Remove Event', 'width=400,height=250,left='+((screen.width-400)/2)+',top='+((screen.height-515)/2)); return false;">
+                    Remove Event
+                    </a>
+                {% else %}
+                    <a href="/game-events/join-event/{{ id }}/"
+                    class="round-button"
+                    onclick="window.open(this.href, 'Join Event', 'width=400,height=250,left='+((screen.width-400)/2)+',top='+((screen.height-515)/2)); return false;">
+                    Join Event
+                    </a>
+                {% endif %}
+            </div>
+       {% else %} 
+        <div>
+            <h1> Event Does not exist. </h1>
+        </div>
+       {% endif %}
+  </body>
+</html>
+{%endblock%}
--- a/game_event/tests.py
+++ b/game_event/tests.py
@@ -1,14 +1,17 @@
 from .models import GameEvent
 from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist
 from court.models import Court
 from decimal import Decimal
 from django.utils import timezone
-from player.models import BallGame
+from player.models import BallGame, Player
 from django.urls import reverse
 from .views import process_game_event_form
 from court_ball_game.models import CourtBallGame
+from game_event_player.models import GameEventPlayer
 from datetime import timedelta
 import pytest
+from django.db import IntegrityError
 
 TEST_ID = 2
 TEST_TIME = timezone.now()
@@ -18,24 +21,39 @@ TEST_MAX = 5
 TEST_BALL_GAME = 'Basketball'
 
 
+@pytest.fixture
+def saved_court():
+    court = Court.objects.create(x=Decimal('11'), y=Decimal('22'),
+                                 city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
+    return court
+
+
+@pytest.fixture
+def saved_game_event(saved_court):
+    return GameEvent.objects.create(id=TEST_ID, time=TEST_TIME, level_of_game=TEST_LEVEL,
+                                    min_number_of_players=TEST_MIN, max_number_of_players=TEST_MAX,
+                                    court=saved_court, ball_game=TEST_BALL_GAME)
+
+
+@pytest.fixture
+def saved_player():
+    player = Player.create('playerr', '0000', '2000-02-12', 'soccer')
+    return player
+
+
+@pytest.fixture
+def saved_game_event_player(saved_game_event, saved_player):
+    return GameEventPlayer.objects.create(game_event=saved_game_event, player=saved_player,
+                                          ball_responsible=False)
+
+
+@pytest.fixture
+def court_ball_game(saved_court):
+    return CourtBallGame.objects.create(court=saved_court, ball_game=TEST_BALL_GAME)
+
+
 @pytest.mark.django_db
 class TestGameEventModel:
-
-    @pytest.fixture
-    def saved_court(self):
-        court = Court.objects.create(x=Decimal('11'), y=Decimal('22'),
-                                     city='TEST_CITY', neighborhood='TEST_NEIGHBORHOOD', max_players=30)
-        return court
-
-    @pytest.fixture
-    def saved_game_event(self, saved_court):
-        return GameEvent.objects.create(id=TEST_ID, time=TEST_TIME, level_of_game=TEST_LEVEL,
-                                        min_number_of_players=TEST_MIN, max_number_of_players=TEST_MAX,
-                                        court=saved_court, ball_game=TEST_BALL_GAME)
-
-    @pytest.fixture
-    def court_ball_game(self, saved_court):
-        return CourtBallGame.objects.create(court=saved_court, ball_game=TEST_BALL_GAME)
 
     def test_get_game_event(self, saved_game_event):
         game_event = GameEvent.objects.get(pk=saved_game_event.pk)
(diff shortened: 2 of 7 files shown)


amitKaplansky:
You can see how the site looks like for now
צילום מסך 2023-05-14 ב-14 17 07
צילום מסך 2023-05-14 ב-14 17 27

צילום מסך 2023-05-14 ב-14 17 42

צילום מסך 2023-05-14 ב-14 17 32
-4a17-a96b-647cfa5f0109)
-4593-979c-dab3a238af72)

game_event/tests.py
```diff
@@ -75,3 +75,27 @@ def test_create_game_event_with_null_ball_game(self, saved_court):
             GameEvent.objects.create(id=TEST_ID, time=TEST_TIME, level_of_game=TEST_LEVEL,
                                      min_number_of_players=TEST_MIN, max_number_of_players=TEST_MAX,
                                      court=saved_court,).full_clean()
+                                     
+
+@pytest.mark.django_db
+class TestGameEventViews:
```
amitCohen2, line 81: I think you also need to add an test 'test_delete_game_event'.
In this case - you need to create an game event, remove it, and try to go again to this url and get a response 404.

game_event/tests.py
```diff
@@ -75,3 +75,27 @@ def test_create_game_event_with_null_ball_game(self, saved_court):
             GameEvent.objects.create(id=TEST_ID, time=TEST_TIME, level_of_game=TEST_LEVEL,
                                      min_number_of_players=TEST_MIN, max_number_of_players=TEST_MAX,
                                      court=saved_court,).full_clean()
+                                     
+
+@pytest.mark.django_db
```
amitCohen2, line 80: I think you also need to add an test 'test_remove_landing_site_game_event'.
In this case - you need to create an game event, remove it, and try to go again to this url and get a response 404.

amitKaplansky: I added the test

amitCohen2:
Looks good at all!
Please fix the PyTest fail for merging as soon as possible

Danielsio (approved):

Danielsio:
Have to say, your PR is full with logic, so first of all Good Job !!
now i think the game-event.html can have a context of the game-event-players, and you should print out the participants as well.
this way you will know if event has some participants that brings the ball with them or not.. 

I think it's important, and not so complicated to implement, what do you say ?

game_event/tests.py
```diff
@@ -179,3 +198,86 @@ def test_post_game_event(self, client, saved_court, court_ball_game):
...
+class TestGameEventViews:
+
+    def test_success_loading_site_game_event(self, client, saved_game_event, player):
+        client.force_login(player.user)
+        game_event_id = saved_game_event.id
+        url = reverse('game_event', args=[game_event_id])
+        response = client.get(url)
+        assert response.status_code == 200
+
+    def test_fail_loading_site_game_event(self, client, player):
```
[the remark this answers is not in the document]
amitKaplansky: If the game event id doesn't exist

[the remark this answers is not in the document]
MaayanMashhadi: > Is there a scenario in which there will be a failure in loading site game event, without getting response status code 404 ? @amitKaplansky @MaayanMashhadi

I think no, because if you look at the view function (game_event), you will see
that if there isn't game_event with specific id mentioned (on this test: -1) the response status code will be 404.
`event = get_object_or_404(GameEvent, pk=id)`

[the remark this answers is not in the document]
OhadSeg: @MaayanMashhadi @amitKaplansky Oh I see, I thought about the possibility will get response status code 200 that everything went well with the response, but logic-wise, there were other problems such as if the game event id doesn't exist, and so the request won't go well even though got status code 200. but considering what maayan stated, it seems fine.

amitKaplansky:
> Have to say, your PR is full with logic, so first of all Good Job !! now i think the game-event.html can have a context of the game-event-players, and you should print out the participants as well. this way you will know if event has some participants that brings the ball with them or not..
> 
> I think it's important, and not so complicated to implement, what do you say ?

You are defiantly right, I added it.

https://github.com/redhat-beyond/MeetBalls/assets/106595737/443c1c89-4272-45f3-aa13-21f7a969517e

game_event/tests.py
```diff
@@ -179,3 +198,86 @@ def test_post_game_event(self, client, saved_court, court_ball_game):
...
+        client.force_login(player.user)
+        response = client.get('/game-events/-1/')
+        assert response.status_code == 404
+
+    def test_success_loading_site_join_game_event(self, client, saved_game_event, player):
```
yftacherzog, line 218: Is this at all different than `test_success_loading_site_game_event`?

game_event/tests.py
```diff
@@ -179,3 +198,86 @@ def test_post_game_event(self, client, saved_court, court_ball_game):
...
+    def test_success_loading_site_remove_game_event(self, client, saved_game_event, saved_game_event_player):
+        client.force_login(saved_game_event_player.player.user)
+        game_event_id = saved_game_event.id
+        url = reverse('remove_from_event', args=[game_event_id])
+        response = client.get(url)
+        assert response.status_code == 200
+
+    def test_fail_loading_site_remove_game_event_from_player(self, client, player):
+        client.force_login(player.user)
+        response = client.get('/game-events/removee-from-event/-1/')
```
yftacherzog, line 239: Is it failing on what you were expecting it to fail on?
```suggestion
        response = client.get('/game-events/remove-from-event/-1/')
```

game_event/tests.py
```diff
@@ -179,3 +198,86 @@ def test_post_game_event(self, client, saved_court, court_ball_game):
...
+
+    def test_remove_landing_site_game_event(self, client, saved_game_event, saved_game_event_player):
+        client.force_login(saved_game_event_player.player.user)
+        game_event_id = saved_game_event.id
+        url = reverse('remove_from_event', args=[game_event_id])
+        response = client.get(url)
+        assert response.status_code == 200
+        saved_game_event.delete()
+        response = client.get(url)
+        assert response.status_code == 404
```
yftacherzog, line 250: Is this part of the functionality you want to test here? is there some added value for deleting the game event and validating its removal? If not, I suggest removing these lines.

game_event/views.py
```diff
@@ -63,3 +65,52 @@ def game_events(request):
...
 
     return render(request, 'game-events.html', {'game_events': game_events})
+
+
+def game_event(request, id):
+    event = get_object_or_404(GameEvent, pk=id)
+    player = Player.objects.get(user=request.user)
+    event_players = GameEventPlayer.objects.filter(game_event=event)
+    event_players = [{'first_name': entry.player.user.first_name, 'last_name': entry.player.user.last_name,
+                     'brings_ball': entry.ball_responsible} for entry in event_players]
```
yftacherzog, line 75: I'd refrain from using the same variable for holding different data types.
Try this instead:
```suggestion
    event_players = [
        {
            "first_name": entry.player.user.first_name,
            "last_name": entry.player.user.last_name,
            "brings_ball": entry.ball_responsible,
        }
        for entry in GameEventPlayer.objects.filter(game_event=event)
    ]
```

game_event/views.py
```diff
@@ -63,3 +65,52 @@ def game_events(request):
...
+    in_event = False
+    if GameEventPlayer.objects.filter(game_event=event, player=player).exists():
+        in_event = True
```
yftacherzog, line 78: Would that work?
```suggestion
    in_event = GameEventPlayer.objects.filter(game_event=event, player=player).exists()
```

game_event/views.py
```diff
@@ -63,3 +65,52 @@ def game_events(request):
...
+    context = {'id': event.id, 'time': event.time, 'level_of_game': event.level_of_game, 'min_number_of_players':
+               event.min_number_of_players, 'max_number_of_players': event.max_number_of_players,
+               'court': event.court.city, 'neighborhood': event.court.neighborhood,
+               'ball_game': event.ball_game, 'in_event': in_event, 'event_players': event_players}
```
yftacherzog, line 82: More readable:
```suggestion
    context = {
        "id": event.id,
        "time": event.time,
        "level_of_game": event.level_of_game,
        "min_number_of_players": event.min_number_of_players,
        "max_number_of_players": event.max_number_of_players,
        "court": event.court.city,
        "neighborhood": event.court.neighborhood,
        "ball_game": event.ball_game,
        "in_event": in_event,
        "event_players": event_players,
    }
```

game_event/views.py
```diff
@@ -63,3 +65,52 @@ def game_events(request):
...
+
+    return render(request, 'game_event/game-event.html', context)
+
+
+def join_event(request, id):
+    get_object_or_404(GameEvent, pk=id)
```
yftacherzog, line 88: Is that for checking that the event exists?

amitKaplansky: yes

yftacherzog: What will happen if you don't do it?

game_event/views.py
```diff
@@ -63,3 +65,52 @@ def game_events(request):
...
+        p = GameEventPlayer.objects.create(game_event=event, player=player, ball_responsible=True)
+        p.save()
+
+    else:
+        p = GameEventPlayer.objects.create(game_event=event, player=player, ball_responsible=False)
+        p.save()
+
+    in_event = False
+    if GameEventPlayer.objects.filter(game_event=event, player=player).exists():
+        in_event = True
```
yftacherzog, line 114:
```suggestion
    in_event = GameEventPlayer.objects.filter(game_event=event, player=player).exists()
```

game_event/views.py
amitKaplansky: I will need to use instead 'GameEvent.objects.get(id=id)' , but after I need to check that the event exists.
'get_object_or_404(GameEvent, pk=id)' does it for me.

game_event/urls.py
```diff
@@ -5,4 +5,8 @@
 urlpatterns = [
     path('create/', views.display_game_event_form, name='display game event form'),
     path('create/process', views.process_game_event_form, name='process game event form'),
-    path('', views.game_events, name='game-events')]
+    path('', views.game_events, name='game-events'),
+    path('<id>/', views.game_event, name='game_event'),
+    path('join-event/<id>/', views.join_event, name='join_event'),
+    path('process-answer/<id>/', views.process_answer, name='process_answer'),
```
yftacherzog, line 11: I think we need a more informative name than process_answer. It's not self explanatory enough (which answer?). Also, suppose tomorrow there will be yet another question, we'd need to somehow differentiate those (`process_answer_2` will it be? :) )

game_event/templates/game_event/game-event.html
```diff
@@ -0,0 +1,67 @@
...
+{% load static %}
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Game Event</title>
+    <link rel="stylesheet" type="text/css" href="{% static 'css/game-event.css' %}">
+    <script>
+        setTimeout(function() {
+            location.reload();
+        }, 5000); // Refresh after 5 seconds (3000 milliseconds)
```
yftacherzog, line 12: I think you can remove the comment

game_event/views.py
yftacherzog: What will happen if you try to join an event that doesn't exist?

The code smells suspicious for two main reasons:
1. You have a `get` call but you don't use the result.
2. If it's just for validating then you have a race condition here: what happens if the event is removed between your check and the `render` call? You are then left with the same situation as you were without having the check.

It's sometimes counter-intuitive, but in most cases it's safer to just try to do what you're trying to do, and process the failure later, rather than checking beforehand that you should be able to do it, as things can change between the check and the doing (asking for forgiveness instead of asking for permission).

Practically, it means putting your doing inside a `try` block and processing the failures inside an `except` block.

game_event/views.py
amitKaplansky: Yes it works, changing it.

game_event/tests.py
amitKaplansky: that's exactly the the functionality we wanted, @amitCohen2 suggested to add this test.

game_event/views.py
amitKaplansky: more readable also, I'll change it as requested.

game_event/views.py
amitKaplansky: I didn't think I could get into a race condition, now I get the point.
by adding the ```try``` ```except``` how will it resolve the race condition?

game_event/tests.py
yftacherzog: ok. It's not really clear to me what is being tested on this test case.

game_event/tests.py
amitKaplansky: No, I'll change it.

game_event/views.py
yftacherzog: The race is between the time that you check the existence and the time that you try to access. If you don't check it exists and just try to access it then there's nothing to race against. It will either succeed or fail. If it fail, you address it on the except block.

game_event/tests.py
amitKaplansky: Amit requested to add a test to check whether you can reach the remove_from_event page after the game event has been deleted. should I change the name of the test?

yftacherzog: ok. I think I understand now. Yes. I think you should rename the test.

game_event/views.py
yftacherzog: You should address it in the exact same way that you would before.

amitKaplansky: do I add empty content if I don't have a game event?

yftacherzog: Up to you. Just make it coherent.

game_event/views.py
```diff
@@ -63,3 +65,52 @@ def game_events(request):
...
+    player_in_event = GameEventPlayer.objects.get(game_event=event, player=player)
+    player_in_event.delete()
+    return render(request, 'game_event/result.html')
+
+
+def process_answer(request, id):
+    event = get_object_or_404(GameEvent, pk=id)
+    player = Player.objects.get(user=request.user)
+    selected_option = request.POST.get('answer')
+    if selected_option == 'yes':
```
Yuval-Vino, line 104: I think it would be easier to understand if you don't write the create method twice.
```

ball_responsible = selected_option == 'yes'
GameEventPlayer.objects.create(game_event=event, player=player, ball_responsible=ball_responsible)

```

game_event/views.py
```diff
@@ -1,6 +1,8 @@
 from .models import GameEvent
-from django.shortcuts import render, redirect
+from django.shortcuts import render, redirect, get_object_or_404
 from court.models import Court
+from game_event_player.models import GameEventPlayer
+from player.models import Player
```
Yuval-Vino, line 5: Please add the import statement for Player within the same line as BallGame instead of creating a new line in the imports.

game_event/tests.py
```diff
@@ -1,14 +1,18 @@
 from .models import GameEvent
 from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist
```
Yuval-Vino, line 3: same here

game_event/tests.py
```diff
@@ -179,3 +198,86 @@ def test_post_game_event(self, client, saved_court, court_ball_game):
...
+            GameEventPlayer.objects.get(game_event=saved_game_event, player=saved_game_event_player.player)
+
+    def test_remove_player_that_is_not_in_event(self, client, saved_game_event, player):
+        client.force_login(player.user)
+        game_event_id = saved_game_event.id
+        url = reverse('remove_from_event', args=[game_event_id])
+        with pytest.raises(ObjectDoesNotExist):
+            client.get(url)
+
+    def test_add_player_to_event(self, client, saved_game_event, player):
```
Yuval-Vino, line 268: Shouldn't we check if a relevant GameEventPlayer has been created?

game_event/tests.py
```diff
@@ -179,3 +198,86 @@ def test_post_game_event(self, client, saved_court, court_ball_game):
...
+        client.force_login(player.user)
+        game_event_id = saved_game_event.id
+        url = reverse('process_answer', args=[game_event_id])
+        response = client.post(url, {'answer': 'yes'})
+        assert response.status_code == 200
+        assert 'in_event' in response.context
```
Yuval-Vino, line 274: why do we need to check that `in_event `is in? Don't we want to check the value of `in_event`?

Yuval-Vino (commented):
Hi @amitKaplansky HTML Looking super good!
i left few comments

game_event/tests.py
amitKaplansky: yes, you defiantly right, I changed it and added of GameEventPlayer has been created.

game_event/views.py
amitKaplansky: The argument in_event is used for the result.html

yftacherzog (approved):

Danielsio (approved):

yftacherzog (approved):
lgtm. make sure you're rebased to prevent last minute issues.

game_event/tests.py
```diff
@@ -179,3 +197,92 @@ def test_post_game_event(self, client, saved_court, court_ball_game):
...
+            client.get(url)
+
+    def test_add_player_to_event(self, client, saved_game_event, player):
+        client.force_login(player.user)
+        game_event_id = saved_game_event.id
+        url = reverse('process_answer_game_event', args=[game_event_id])
+        response = client.post(url, {'answer': 'yes'})
+        assert response.status_code == 200
+        assert response.context['in_event'] is True
+        assert GameEventPlayer.objects.filter(game_event=saved_game_event, player=player).exists() is True
```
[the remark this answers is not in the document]
amitKaplansky: it was more readable for me, but ill do it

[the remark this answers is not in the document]
Yuval-Vino: if it is more readable for you, then let's keep it as is,
but why is assert GameEventPlayer.exists() not readable?

Yuval-Vino (approved):

Yuval-Vino (approved):

PR #90, opened 2023-05-24
Add Notification Entity

### LINKED ISSUE #88: Add Notification Entity
**Expected Behavior**
A user will be able to see notifications about own events

**Current Behavior**
no notifications page.

amitCohen2:
### Desired Outcome

1.  Users as players would have a notifications.
2. This entity is required for have this notification page 

### Implemented Changes

Add notification/ file which include all the entity required files:
1. models.py
2. tests.py 
3. admin.py
4. migrations
5. veiws.py - Should be empty

Add to conftest.py notification fixuse.

### Connected Issue/Story

Resolves #88

### Dependencies

### Definition of Done
*Feel free to copy this information from the connected issue.*

#### Test coverage

  changes

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: [89]()


notification/models.py
amitKaplansky, line 1: The field "player_id" should be renamed to "player" to avoid confusion, it will clarify that the field refers to the player itself, rather than just the id of player.

amitKaplansky: why do you need seen time? and by who?

amitKaplansky: You should modify the player class by overriding the str method, the default implementation returns a string that includes the class name and the memory address of the object, and I don't think this is what you want

notification/tests.py
amitKaplansky, line 1: we already have player fixture in the contest I don't think you need to add it.

notification/migrations/0001_initial.py
```diff
@@ -0,0 +1,34 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Notification',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('seen_time', models.DateTimeField(blank=True, null=True)),
+                ('sent_time', models.DateTimeField(default=django.utils.timezone.now)),
+                ('message', models.TextField(blank=True, null=True)),
```
MaayanMashhadi, line 23: Can you explain me the purpose of this field? If I understand correctly this line, you created 'message' text field. Consider maybe to use the message model we have already created, because his purpose is to store text message,
but maybe I don't understand the purpose of this field here.

notification/migrations/0001_initial.py
```diff
@@ -0,0 +1,34 @@
...
+        ('player', '0002_static_players'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Notification',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('seen_time', models.DateTimeField(blank=True, null=True)),
+                ('sent_time', models.DateTimeField(default=django.utils.timezone.now)),
```
MaayanMashhadi, line 22: There is a 'time_sent' field in message model. with this and the previous review (about
the 'meesage' field) consider to use the message model here (I think it feels that there is some
duplicating here).

notification/migrations/0001_initial.py
```diff
@@ -0,0 +1,34 @@
...
+                ('message', models.TextField(blank=True, null=True)),
+                ('notification_type', models.CharField(choices=[('website', 'Website'), ('game-event', 'Game-Event')],
```
MaayanMashhadi, line 24: As I saw the other fields here, I assume that you want to use the notification in the chat (correct me if I wrong).
If so, you don't need here type of 'chat'?

notification/tests.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_create_notification(self, notification):
+        assert isinstance(notification, Notification)
+        assert notification.player_id is not None
+        assert notification.sent_time is not None
+        assert notification.message == 'Test notification'
+        assert notification.notification_type == NotificationType.WEBSITE
+        assert notification.is_read is False
+
+    def test_persisted_notification_in_database(self, notification):
+        assert notification in Notification.objects.all()
```
MaayanMashhadi, line 40: Can you explain me the purpose of this test? Do you want to check if the database acts normally?

notification/tests.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_delete_notification_from_database(self, notification):
+        notification.delete()
```
MaayanMashhadi, line 43: Same here

MaayanMashhadi (commented):
Hi Amit, nice job!
I left you some reviews, I would like to hear your opinion (I think I don't really understand part of the fields model, I would like you to explain me)

notification/migrations/0001_initial.py
amitCohen2: This shopuld be an text field for the notification.
I dont understand ehat are you asking me. 
Maybe to use the message model we have instead this field?

MaayanMashhadi: > This should be a text field for the notification. I dont understand what are you asking me. Maybe to use the message model we have instead this field?

Ok, now I'm understanding what you meant.
You don't need to do what I said, it's great like that.

notification/tests.py
amitCohen2: Yep.
The purpose of this test is to verify if a notification object is correctly persisted in a database.

notification/tests.py
amitCohen2: Same like above, 
The purpose of this test is to delete a notification object from the database.

notification/tests.py
amitCohen2: Thanks, I'll check it out

notification/migrations/0001_initial.py
amitCohen2: Added, thanks for notice that

notification/tests.py
amitCohen2: Fixed this

notification/migrations/0001_initial.py
amitCohen2: Ok, but if you see some things like that its good.
Maybe its better to make a connection between messeges and notfication or seperate them like that.
Now I think its good the make it like that seperatly.

notification/models.py
amitCohen2: > The field "player_id" should be renamed to "player" to avoid confusion, it will clarify that the field refers to the player itself, rather than just the id of player.

Changed
> why do you need seen time? and by who?

Thanks for notice me that, removed

> You should modify the player class by overriding the str method, the default implementation returns a string that includes the class name and the memory address of the object, and I don't think this is what you want

Can you explain me this in other words? I dont undestant whats your point.

notification/models.py
```diff
@@ -0,0 +1,48 @@
...
+        default=NotificationType.WEBSITE,
+    )
+    is_read = models.BooleanField(default=False)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=['player_id', 'sent_time'], name='unique notification')
+        ]
+
+    def save(self, *args, **kwargs):
```
Yuval-Vino, line 28: Could you please explain the purpose behind your changes? From what I observe, you have mainly rewritten the `super.save` method and made slight modifications to the error messages. Typically, when rewriting the save function, I would anticipate additional input validations or other modifications that are not present in the original save method. Can you clarify the intention and any additional functionalities you aimed to incorporate with your changes?

notification/models.py
amitKaplansky: If you don't override the __str__ method, the default implementation returns a string that includes the class name and the memory address of the object, I think that is not what you mean. @Yuval-Vino added that it would be more suitable to put there `self.notification_type`

notification/tests.py
amitKaplansky: Can you please also add `notification` fixture to conftest?

notification/models.py
```diff
@@ -0,0 +1,48 @@
...
+    notification_type = models.CharField(
+        max_length=100,
+        choices=NotificationType.choices,
+        default=NotificationType.WEBSITE,
+    )
+    is_read = models.BooleanField(default=False)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=['player_id', 'sent_time'], name='unique notification')
```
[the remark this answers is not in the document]
amitKaplansky: maybe we should also add a DateTime field `relevant_time` or other more meaningful name, if the player has not seen the message and its no longer relevant there isn't a need to present the message any more.
what do you say?

notification/models.py
```diff
@@ -0,0 +1,48 @@
+from django.db import models, IntegrityError
+from django.utils import timezone
+from player.models import Player
+
+
+class NotificationType(models.TextChoices):
+    WEBSITE = "website", "Website"
+    EVENT = "game-event", "Game-Event"
+    CHAT = "chat", "Chat"
```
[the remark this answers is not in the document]
amitCohen2: Sure.
EVENT - for event notifications
CHAT - for future chat notification
WEBSITE - for event recommendations according to your favorite ball game

notification/models.py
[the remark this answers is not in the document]
amitCohen2: > is this constraint necessary? what if 2 different events that the player is registered for have changed at the same time and both want to send notification to the same player at the same time

I didn't think on that case. Remove this soon.

> maybe we should also add a DateTime field `relevant_time` or other more meaningful name, if the player has not seen the message and its no longer relevant there isn't a need to present the message any more. what do you say?

I don't think this `relevant time`  field is necessary. 
Try to look on this model and see if we need more connections to other relevant entities. This eill be more necessary.

notification/models.py
amitCohen2: I just eanted to assert if we have an 'UniqueConstraint' problem.
I saw now its not necessary at all because we can have two different events that the player is registered for have changed at the same time.
Will Remove this save method.

notification/models.py
amitCohen2: > If you don't override the **str** method, the default implementation returns a string that includes the class name and the memory address of the object, I think that is not what you mean. @Yuval-Vino added that it would be more suitable to put there `self.notification_type`

Thanks foe explain me this, Fixed!

notification/tests.py
```diff
@@ -0,0 +1,54 @@
...
+            Notification.objects.create(
+                player=player,
+                sent_time=notification1.sent_time,
+                message='Test notification 2',
+                notification_type=NotificationType.WEBSITE,
+                is_read=False
+            )
+
+    def test_mark_as_read_notification(self, notification):
+        updated_notification = Notification.objects.get(id=notification.id)
```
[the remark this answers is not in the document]
amitCohen2: Fixed this test

notification/models.py
```diff
@@ -0,0 +1,48 @@
...
+        try:
+            result = super().save(*args, **kwargs)
+        except IntegrityError as error:
+            if row_duplication_error in error.args:
+                raise IntegrityError("notification already exists")
+            raise error
+        return result
+
+    def __str__(self):
+        return f"{self.player} - {self.message}"
```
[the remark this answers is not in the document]
amitCohen2: Thanks, Amit K. also told me about it.
Fixed

notification/tests.py
```diff
@@ -0,0 +1,54 @@
...
+        message='Test notification',
+        notification_type=NotificationType.WEBSITE,
+        is_read=False
+    )
+
+
+@pytest.mark.django_db
+class TestNotification:
+    def test_create_notification(self, notification):
+        assert isinstance(notification, Notification)
```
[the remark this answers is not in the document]
amitCohen2: Removed the other asserts

notification/tests.py
```diff
@@ -0,0 +1,54 @@
...
+        assert notification.message == 'Test notification'
+        assert notification.notification_type == NotificationType.WEBSITE
+        assert notification.is_read is False
+
+    def test_persisted_notification_in_database(self, notification):
+        assert notification in Notification.objects.all()
+
+    def test_delete_notification_from_database(self, notification):
+        notification.delete()
+        assert notification not in Notification.objects.all()
```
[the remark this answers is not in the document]
amitCohen2: Thanks for explain me this, 
Fixed!

notification/models.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @staticmethod
+    def mark_notification_as_read(notification_id):
```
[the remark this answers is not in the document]
amitCohen2: Fixed this,  the PyCharm offer me this shitty solution ;)

notification/models.py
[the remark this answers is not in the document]
Yuval-Vino: you need to remigrate after those changes

notification/tests.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_persisted_notification_in_database(self, notification):
+        assert notification in Notification.objects.all()
+
+    def test_delete_notification_from_database(self, notification):
+        notification.delete()
+        with pytest.raises(Notification.DoesNotExist):
+            Notification.objects.get(pk=notification.pk)
+
+    def test_mark_as_read_notification(self, notification):
+        Notification.mark_notification_as_read(notification)
```
Yuval-Vino, line 31: After making this function non-static, you can enjoy the benefits it offers
```suggestion
        notification.mark_notification_as_read()
```

notification/models.py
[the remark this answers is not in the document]
amitCohen2: Done

notification/tests.py
amitCohen2: Thanks for this tip, 
Done

Danielsio (approved):
Good Work Amit ! your tests looks sufficiant to me, I have no comments to add.

notification/tests.py
```diff
@@ -0,0 +1,20 @@
...
+        assert notification in Notification.objects.all()
+
+    def test_delete_notification_from_database(self, notification):
+        notification.delete()
+        with pytest.raises(Notification.DoesNotExist):
+            Notification.objects.get(pk=notification.pk)
+
+    def test_mark_as_read_notification(self, notification):
+        notification.mark_notification_as_read()
+        assert notification.is_read is True
```
yftacherzog, line 20: I think we better check it was false before to see that we made a difference.

amitCohen2: Added this check line to this test.

yftacherzog:
I see the PR description mentions the ERD, but it's not being updated here. I suggest either removing this from the description or having it implemented. It also mentions the views file. If it's suppose to be empty, I think it's better to mention it.

amitCohen2:
> I see the PR description mentions the ERD, but it's not being updated here. I suggest either removing this from the description or having it implemented. It also mentions the views file. If it's suppose to be empty, I think it's better to mention it.

Thanks for notice me that.
We will update the ERD in other PR. Removed from the description.

yftacherzog (approved):

PR #94, opened 2023-05-25
Add current weather info to game event table

### LINKED ISSUE #91: Add current weather to game events table
## Expected Behavior
Users could see the current weather in the city where the court of the game event is located

## Current Behavior
There is currently no implementation for it 

### Acceptance Criteria
1. Set an API key for a weather api
2. Modify view - add a get request to the weather api to get the weather
3. Change the template for the game events table
4. add tests

OhadSeg:
### Desired Outcome

Current temperature and weather icons will be added to the game event table for each game event.

### Implemented Changes

Modify game events template, view, and tests. Current weather details are shown in the game event table.

### Connected Issue/Story
Resolves #91

### Dependencies

This section inidicates whether this PR depends on any other PR's. 
Might need to wait for other PR's to merged before. Add the PR ID.
- [x] This PR depends on other PR's
  changes
- [ ] This PR does not depend on other PR's 

### Definition of Done

1.Add view that make http get request to openweather api at game_event/views.py 
2.Modify game-events.html template
3.Add tests for the weather view at game_event/tests.py

#### Test coverage

  changes
- [ ] The changes in this PR do not require tests

#### Documentation

- [ ] Docs (e.g. `README`s) were updated in this PR
- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID]()

image

### DIFF
--- a/game_event/templates/game-events.html
+++ b/game_event/templates/game-events.html
@@ -36,6 +36,9 @@
 				<th>Minimum Players</th>
 				<th>Maximum Players</th>
 				<th>Court ID</th>
+                <th>City</th>
+                <th>Curr Temp</th>
+                <th>Curr Weather</th>
 				<th>Ballgame</th>
 			</tr>
 		</thead>
@@ -44,12 +47,20 @@
 		<tbody>
 			{% for event in game_events %}
 			<tr>
-				<td><a href="http://127.0.0.1:8000/game-events/{{ event.id }}/">{{ event.id }}</a></td>
+				<td><a href="/game-events/{{ event.id }}/">{{ event.id }}</a></td>
 				<td>{{ event.time }}</td>
 				<td>{{ event.level_of_game }}</td>
 				<td>{{ event.min_number_of_players }}</td>
 				<td>{{ event.max_number_of_players }}</td>
-				<td><a href="http://127.0.0.1:8000/courts/{{ event.court.courtID }}">{{ event.court.courtID }}</a></td>
+				<td><a href="/courts/{{ event.court.courtID }}">{{ event.court.courtID }}</a></td>
+                <td>{{ event.court.city }}</td>
+                {% if event.weather %}
+                    <td>  {{ event.weather.temp}} </td>
+                    <td>  <img src={{ event.weather.icon }}> </td>
+                {% else %}
+                    <td> N/A </td>
+                    <td> N/A </td>
+                {% endif %}
 				<td>{{ event.ball_game }}</td>
 			</tr>
 			{% endfor %}
--- a/game_event/test_game_event_views.py
+++ b/game_event/test_game_event_views.py
@@ -17,29 +17,35 @@ class TestGameEventServerResponses:
     def test_shown_min_player_filter(self, client, game_event):
         response = client.get(reverse(game_events), {'min_players': '2'})
         assert response.status_code == 200
-        assert game_event in response.context['game_events']
+        game_events_response = response.context['game_events']
+        assert any(game.id == game_event.id for game in game_events_response)
 
     def test_not_shown_min_player_filter(self, client, game_event):
         response = client.get(reverse(game_events), {'min_players': '10'})
         assert response.status_code == 200
-        assert game_event not in response.context['game_events']
+        game_events_response = response.context['game_events']
+        assert not any(game.id == game_event.id for game in game_events_response)
 
     def test_shown_max_player_filter(self, client, game_event):
         response = client.get(reverse(game_events), {'max_players': '15'})
         assert response.status_code == 200
-        assert game_event in response.context['game_events']
+        game_events_response = response.context['game_events']
+        assert any(game.id == game_event.id for game in game_events_response)
 
     def test_not_shown_max_player_filter(self, client, game_event):
         response = client.get(reverse(game_events), {'max_players': '2'})
         assert response.status_code == 200
-        assert game_event not in response.context['game_events']
+        game_events_response = response.context['game_events']
+        assert not any(game.id == game_event.id for game in game_events_response)
 
     def test_shown_max_level_filter(self, client, game_event):
         response = client.get(reverse(game_events), {'max_level': '10'})
         assert response.status_code == 200
-        assert game_event in response.context['game_events']
+        game_events_response = response.context['game_events']
+        assert any(game.id == game_event.id for game in game_events_response)
 
     def test_not_shown_max_level_filter(self, client, game_event):
         response = client.get(reverse(game_events), {'max_level': '2'})
         assert response.status_code == 200
-        assert game_event not in response.context['game_events']
+        game_events_response = response.context['game_events']
+        assert not any(game.id == game_event.id for game in game_events_response)
--- a/game_event/tests.py
+++ b/game_event/tests.py
@@ -5,10 +5,13 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.utils import timezone
 from player.models import BallGame
 from django.urls import reverse
-from .views import process_game_event_form
+from .views import process_game_event_form, get_weather_from_api
 from datetime import timedelta
 import pytest
 from django.db import IntegrityError
+from unittest import mock
+import urllib
+
 
 TEST_ID = 2
 TEST_TIME = timezone.now()
@@ -16,6 +19,24 @@ TEST_LEVEL = 3
 TEST_MIN = 2
 TEST_MAX = 5
 TEST_BALL_GAME = 'Basketball'
+
+
+@pytest.fixture
+def mock_urlopen():
+    return mock.create_autospec(urllib.request.urlopen)
+
+
+@pytest.fixture
+def another_game_event(court):
+    return GameEvent.objects.create(
+        id=1002,
+        time=timezone.now() - timezone.timedelta(days=1),
+        level_of_game=5,
+        min_number_of_players=4,
+        max_number_of_players=10,
+        court=court,
+        ball_game='Basketball'
+    )
 
 
 @pytest.mark.django_db
@@ -252,3 +273,69 @@ class TestGameEventViews:
         assert response.status_code == 200
         with pytest.raises(IntegrityError):
             response = client.post(url, {'answer': 'yes'})
+
+
+@pytest.mark.django_db
+class TestGetWeatherFromApi:
+    def test_one_valid_response(self, game_event, player, mock_urlopen, court):
+        mock_response = mock.Mock()
+        mock_response.status = 200
+        mock_response.read.return_value = bytes('{"weather": [{"icon": "04d"}],"main": {"temp": 13.19}}', 'utf-8')
+        mock_urlopen.return_value = mock_response
+
+        assert not hasattr(game_event, 'weather')
+        game_events = [game_event]
+
+        size_before = len(game_events)
+        output = get_weather_from_api(game_events, mock_urlopen)
+        assert size_before == len(output)
+
+        expected_weather = {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
+
+        weathers = [game.weather for game in output]
+        assert weathers == [expected_weather]
+
+    def test_exception_thrown_from_api_call(self, game_event, player, mock_urlopen):
+        mock_urlopen.side_effect = Exception()
+
+        assert not hasattr(game_event, 'weather')
+        game_events = [game_event]
+
+        size_before = len(game_events)
+        output = get_weather_from_api(game_events, mock_urlopen)
+        assert size_before == len(output)
+
+        weathers = [game.weather for game in output]
+        assert weathers == [None]
+
+    def test_one_invalid_response_status_code(self, game_event, player, mock_urlopen):
+        mock_response = mock.Mock()
+        mock_response.status = 185
+        mock_urlopen.return_value = mock_response
+
+        assert not hasattr(game_event, 'weather')
+        game_events = [game_event]
+
+        size_before = len(game_events)
+        output = get_weather_from_api(game_events, mock_urlopen)
+        assert size_before == len(output)
+
+        weathers = [game.weather for game in output]
+        assert weathers == [None]
+
+    def test_empty_game_events_list_valid_empty_dictionary_response(self, player):
+        game_events = []
+        get_weather_from_api(game_events, "does not matter")
+        assert game_events == []
+
+    def test_two_valid_game_events_list_valid_response(self, game_event, another_game_event, player, mock_urlopen):
+        mock_response = mock.Mock()
+        mock_response.status = 200
+        mock_response.read.return_value = bytes('{"weather": [{"icon": "04d"}],"main": {"temp": 13.19}}', 'utf-8')
+        mock_urlopen.return_value = mock_response
+        game_events = [game_event, another_game_event]
+        output = get_weather_from_api(game_events, mock_urlopen)
+        expected_weather = {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
+
+        weathers = [game.weather for game in output]
+        assert weathers == [expected_weather, expected_weather]
--- a/game_event/views.py
+++ b/game_event/views.py
@@ -9,6 +9,9 @@ from django.core.exceptions import ValidationError
 from django.contrib import messages
 from message.models import Message
 from datetime import date
+import urllib.request
+import json
+import os
 
 
 def display_game_event_form(request):
@@ -47,7 +50,6 @@ def process_game_event_form(request):
 
 def game_events(request):
     game_events = GameEvent.objects.all()
-
     if request.GET.get('min_players'):
         min_players = int(request.GET.get('min_players'))
         game_events = game_events.filter(min_number_of_players__gte=min_players)
@@ -63,6 +65,8 @@ def game_events(request):
     if request.GET.get('hide_past_events'):
         current_date = date.today()
         game_events = game_events.filter(time__gte=current_date)
+
+    game_events = get_weather_from_api(game_events, urllib.request.urlopen)
 
     return render(request, 'game-events.html', {'game_events': game_events})
 
@@ -153,3 +157,30 @@ def process_answer_game_event(request, id):
     content['event_exists'] = True
 
     return render(request, 'game_event/result.html', content)
+
+
+def get_weather_from_api(game_events, weather_getter):
+    url = 'https://api.openweathermap.org/data/2.5/weather'
+    params = {'units': 'metric', 'appid': os.getenv("APPID")}
+    for game in game_events:
+
+        params['q'] = game.court.city
+
+        full_url = url + '?' + urllib.parse.urlencode(params)
+        try:
+            response = weather_getter(full_url)
+        except Exception:
+            game.weather = None
+            continue
+
+        if response.status == 200:
+            data = response.read().decode('utf-8')
+            weather_data = json.loads(data)
+            curr_icon = weather_data['weather'][0]['icon']
+            icon_url = "http://openweathermap.org/img/wn/" + curr_icon + "@2x.png"
+            curr_temp = weather_data['main']['temp']
+            game.weather = {'icon': icon_url, 'temp': curr_temp}
+        else:
+            game.weather = None
+
+    return game_events


game_event/views.py
```diff
@@ -63,7 +65,21 @@ def game_events(request):
...
+            'time': game.time,
+            'level_of_game': game.level_of_game,
+            'min_number_of_players': game.min_number_of_players,
+            'max_number_of_players': game.max_number_of_players,
+            'court': game.court,
+            'ball_game': game.ball_game,
+            'weather': weather[game.id]
+        })
+
+    return render(request, 'game-events.html', {'game_events': game_events_with_weather})
```
OhadSeg, line 82: @ZvikaNaorCohen I think that when I change here `return render(request, 'game-events.html', {'game_events': game_events_with_weather})` the game_events value, the tests that fail now, are dependent on his value also. 
lets take it 1on1 and have a look.

OhadSeg: @ZvikaNaorCohen Failed tests:
`game_event/test_game_event_views.py::TestGameEventServerResponses::test_shown_min_player_filter FAILED                        
game_event/test_game_event_views.py::TestGameEventServerResponses::test_shown_max_player_filter FAILED                        
game_event/test_game_event_views.py::TestGameEventServerResponses::test_shown_max_level_filter FAILED`

game_event/tests.py
```diff
@@ -286,3 +289,46 @@ def test_fail_add_player_to_event_already_in_it(self, client, saved_game_event,
...
+        url = 'https://api.openweathermap.org/data/2.5/weather'
+        params = {
+            'q': invalid_city,
+            'units': 'metric',
+            'appid': 'bd708048ad9d3ee449660bedad49f30e'
+        }
+        full_url = url + '?' + urllib.parse.urlencode(params)
+        try:
+            response = urllib.request.urlopen(full_url)
+            # assert response.status == 404
```
amitCohen2, line 307: can you please remove this notes?
It implemented in th next block

OhadSeg: Good point, @amitCohen2 Sure, after we'll fix the broken tests, I'll remove all exceeding prints and note

game_event/views.py
OhadSeg: I made changes in all of the tests in this file in accordance with @ZvikaNaorCohen changes.

game_event/tests.py
OhadSeg: Removed the comments when solving broken tests @amitCohen2 .

game_event/test_game_event_views.py
amitKaplansky, line 1: Do the changes in game-events.html require any updates to these tests?

game_event/views.py
amitKaplansky, line 1: is the print on line 69 is needed? can it be removed?

game_event/templates/game-events.html
```diff
@@ -36,6 +36,9 @@ <h1>Game Events</h1>
 				<th>Minimum Players</th>
 				<th>Maximum Players</th>
 				<th>Court ID</th>
+                <th>City</th>
```
SchSeba, line 39: please fix the indentation here

game_event/templates/game-events.html
```diff
@@ -49,6 +52,14 @@ <h1>Game Events</h1>
 				<td>{{ event.min_number_of_players }}</td>
 				<td>{{ event.max_number_of_players }}</td>
 				<td><a href="http://127.0.0.1:8000/courts/{{ event.court.courtID }}">{{ event.court.courtID }}</a></td>
+                <td>{{ event.court.city }}</td>
```
SchSeba, line 55: same here

game_event/templates/game-events.html
```diff
@@ -49,6 +52,14 @@ <h1>Game Events</h1>
 				<td>{{ event.min_number_of_players }}</td>
 				<td>{{ event.max_number_of_players }}</td>
 				<td><a href="http://127.0.0.1:8000/courts/{{ event.court.courtID }}">{{ event.court.courtID }}</a></td>
```
SchSeba, line 54: I know you didn't change this but this is bad it will not work in a "real" deployment.
you can change this to `/courts/` and I think the browser will be able to use the base url

game_event/views.py
```diff
@@ -150,3 +166,34 @@ def process_answer_game_event(request, id):
...
+
+
+def get_weather_from_api(game_events):
+    weather = {}
+    for game in game_events:
+        url = 'https://api.openweathermap.org/data/2.5/weather'
+        params = {
+            'q': game.court.city,
+            'units': 'metric',
+            'appid': 'bd708048ad9d3ee449660bedad49f30e'
```
SchSeba, line 178: be aware that in general exposing the api key I the code is really bad from security point of view.

game_event/views.py
```diff
@@ -150,3 +166,34 @@ def process_answer_game_event(request, id):
...
+        }
+
+        full_url = url + '?' + urllib.parse.urlencode(params)
+        try:
+            response = urllib.request.urlopen(full_url)
+            if response.status == 200:
+                data = response.read().decode('utf-8')
+                print("Response good")
```
SchSeba, line 186: please remove the print debugs

game_event/views.py
```diff
@@ -150,3 +166,34 @@ def process_answer_game_event(request, id):
     content['event_exists'] = True
 
     return render(request, 'game_event/result.html', content)
+
+
+def get_weather_from_api(game_events):
```
SchSeba, line 171: why this function doesn't just return the new list instead of running over again in the view function?

game_event/test_game_event_views.py
OhadSeg: @amitKaplansky Regarding the current changes I have made, I don't think so, as for all the tests are passing on my side. The problem with the tests in the game_event/test_game_event_views.py, was that they were based on the game_event object in the game_event list, created in the matching view file, while I made changes to the game_event objects in the game_events list, by adding on (not in model) weather field.

game_event/views.py
OhadSeg: @OhadSeg yes, It's redundant, will be removed.

game_event/views.py
OhadSeg: Hey, @SchSeba I'm aware of it, I also got a warning from 'GitGuardian', But I don't know what is a good practice in that case. I would use a different static file with all of the app secrets, and also this api key, and add this file to .gitignore file, so it would be available locally to me and wouldn't be exposed. But in this case how would the requests will work for the rest of our team? For example in this case, where I could the appid stored it unshared different file: 
`params = {
            'q': game.court.city,
            'units': 'metric',
            'appid': weatherSecret`

The flake8 and pytest would probably won't work, and if will work, than the http request will fail.

game_event/views.py
OhadSeg: @SchSeba Removed.

game_event/templates/game-events.html
OhadSeg: @SchSeba The indentation in this file is different and OK on my local repo, and I didn't make changes in the template, how is it possible?
image

game_event/templates/game-events.html
OhadSeg: @SchSeba again in this case, looks file on my IDE : 
image

game_event/views.py
IdanZimi: Would be great if someone can elaborate on that issue

OhadSeg: @yftacherzog @Yuval-Vino @SchSeba

game_event/templates/game-events.html
OhadSeg: @SchSeba Chaned the path. Fixed.

game_event/views.py
OhadSeg: @SchSeba I think that even though it is more code in the main view function `def game_events(request)` it is more readable if the final `game_events_with_weather` object that we send when rendering, is shown to the reader in this function, and not at the `def get_weather_from_api(game_events)` function.

game_event/templates/game-events.html
OhadSeg: @SchSeba Fixed.

game_event/templates/game-events.html
SchSeba: weird that is fine then..

game_event/templates/game-events.html
SchSeba: weird that is fine then..

game_event/views.py
SchSeba: I think you can load all the secret you need in the settings file from django by using environment variables.
then you can share with the team what are the env variables needed to run the application

game_event/tests.py
```diff
@@ -286,3 +289,41 @@ def test_fail_add_player_to_event_already_in_it(self, client, saved_game_event,
         assert response.status_code == 200
         with pytest.raises(IntegrityError):
             response = client.post(url, {'answer': 'yes'})
+
+
+@pytest.mark.django_db
+class TestWeatherApi:
```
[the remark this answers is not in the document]
yftacherzog: @OhadSeg 
We shouldn't.
This test case is kind of testing the library rather than the code.

Also, unit tests should not try and interact with external APIs.

The way to go about it is to use the concept of mocks: https://docs.python.org/3/library/unittest.mock.html
It allows you to inject an object that can be used by the code as if it was that library, but in reality, it will return whatever you tell it to return.

You should have a function that does only the call to the API, and you test it by mocking the external library - on each test case you'll mock a different output and verify that the function behaves accordingly.

Then when unit testing functions that use this function, you can mock the function itself rather than the API, so it returns what you need it to return.

It's a bit big. If you want, you can schedule a call and we can look into it together.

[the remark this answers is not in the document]
OhadSeg: After a call with @yftacherzog, @IdanZimi and I removed the explicitly stated API Key, and change it to `'appid': os.getenv("APPID")`, so that when running the server, we should attach the API-KEY ad environment variable, like this : 
`APPID='api-key' pipenv run python manage.py runserver 0.0.0.0:8000`.
we also changed the view function, so if someone run the server without attaching the api key, `pipenv run python manage.py runserver 0.0.0.0:8000` there won't be any problems, just that the game events table, will render without it.

game_event/views.py
OhadSeg: After a call with @yftacherzog, @IdanZimi and I removed the explicitly stated API Key, and change it to 'appid': os.getenv("APPID"), so that when running the server, we should attach the API-KEY ad environment variable, like this :
APPID='api-key' pipenv run python manage.py runserver 0.0.0.0:8000.
we also changed the view function, so if someone run the server without attaching the api key, pipenv run python manage.py runserver 0.0.0.0:8000 there won't be any problems, just that the game events table, will render without it.

game_event/tests.py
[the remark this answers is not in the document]
yftacherzog: The test file has conflicts. You'd need to rebase and resolve those.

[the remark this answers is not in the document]
OhadSeg: @yftacherzog Done.

game_event/tests.py
```diff
@@ -31,6 +34,34 @@ def saved_game_event_player(saved_game_event, player):
...
 
+@pytest.fixture
+def court_ball_game(court):
+    return CourtBallGame.objects.create(court=court, ball_game=TEST_BALL_GAME)
+
+
+@pytest.fixture
+def mock_urlopen():
+    with mock.patch('urllib.request.urlopen') as mock_urlopen:
+        return mock_urlopen
```
yftacherzog, line 45: Patching is the act of replacing something within another object by a mock. In this case, you just want to return the mock, so you don't need to patch.
```suggestion
    return mock.Mock()
```

Creating the mock this way, means that it will accept anything you'd throw its way. e.g. it can run `mock_urlopen.does_not_exist`

You can also create more accurate mocks, which is encouraged, which will only pretend to have the capabilities of the mocked object. You do this with `create_autospec`:
```suggestion
    return mock.create_autospec(urllib.request.urlopen)
```

See the difference here:
```
In [9]: gen_mock = mock.Mock()

In [10]: gen_mock.say_ahh()
Out[10]: <Mock name='mock.say_ahh()' id='140540985628320'>

In [11]: spec_mock = mock.create_autospec(urllib.request.urlopen)

In [12]: spec_mock.say_ahh()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [12], line 1
----> 1 spec_mock.say_ahh()

AttributeError: 'function' object has no attribute 'say_ahh'

In [13]: spec_mock('some-url')
Out[13]: <MagicMock name='mock()' id='140541262404672'>
```

All in all:
- You don't need the patch and the context manager, just return the mock.
- use autospec instead of generic mocks when possible.

game_event/tests.py
```diff
@@ -31,6 +34,34 @@ def saved_game_event_player(saved_game_event, player):
...
+        time=timezone.now() - timezone.timedelta(days=1),
+        level_of_game=5,
+        min_number_of_players=4,
+        max_number_of_players=10,
+        court=court,
+        ball_game='Basketball'
+    )
+    game_events_list.append(game_event)
+    game_events_list.append(saved_game_event)
+    return game_events_list
```
yftacherzog, line 62: It's more readable if you visually put the objects inside a list rather than using append.
```suggestion
    game_event = GameEvent.objects.create(
        id=1001,
        time=timezone.now() - timezone.timedelta(days=1),
        level_of_game=5,
        min_number_of_players=4,
        max_number_of_players=10,
        court=court,
        ball_game='Basketball'
    )
    return [
        game_event,
        saved_game_event,
    ]
```

game_event/tests.py
```diff
@@ -265,3 +296,53 @@ def test_fail_add_player_to_event_already_in_it(self, client, saved_game_event,
         assert response.status_code == 200
         with pytest.raises(IntegrityError):
             response = client.post(url, {'answer': 'yes'})
+
+
+@pytest.mark.django_db
+class TestWeatherApi:
```
yftacherzog, line 302: We're testing our function rather than the API.
```suggestion
class TestGetWeatherFromApi:
```

game_event/views.py
```diff
@@ -150,3 +166,33 @@ def process_answer_game_event(request, id):
...
+        try:
+            response = weather_getter(full_url)
+        except Exception:
+            weather[game.id] = None
+            continue
+
+        if response.status == 200:
+            data = response.read().decode('utf-8')
+            weather_data = json.loads(data)
+            currIcon = weather_data['weather'][0]['icon']
```
yftacherzog, line 191:
```suggestion
            curr_icon = weather_data['weather'][0]['icon']
```

game_event/views.py
```diff
@@ -150,3 +166,33 @@ def process_answer_game_event(request, id):
...
+            iconUrl = "http://openweathermap.org/img/wn/" + currIcon + "@2x.png"
+            currTemp = weather_data['main']['temp']
```
yftacherzog, line 193:
```suggestion
            curr_temp = weather_data['main']['temp']
```

game_event/tests.py
```diff
@@ -265,3 +296,53 @@ def test_fail_add_player_to_event_already_in_it(self, client, saved_game_event,
...
+    def test_one_valid_response(self, client, saved_game_event, player, mock_urlopen):
+        mock_response = mock.Mock()
+        mock_response.status = 200
+        mock_response.read.return_value = bytes('{"weather": [{"icon": "04d"}],"main": {"temp": 13.19}}', 'utf-8')
+        mock_urlopen.return_value = mock_response
+
+        client.force_login(player.user)
```
yftacherzog, line 309: Is this line needed? Will it fail if we just remove it?

yftacherzog: (Same goes for the rest of the tests in this class)

game_event/tests.py
```diff
@@ -265,3 +296,53 @@ def test_fail_add_player_to_event_already_in_it(self, client, saved_game_event,
...
+
+        output = get_weather_from_api([saved_game_event], mock_urlopen)
+        assert output[saved_game_event.id] == {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
```
yftacherzog, line 312: I think the test will be more complete if we assert on the complete output rather than one value (imagine a regression in which the function injects a trailing entry - you won't catch it if you only look at a specific entry):
```suggestion
        expected_weather = {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
        assert output == {
                saved_game_event.id: expected_weather
        }
```

game_event/tests.py
```diff
@@ -265,3 +296,53 @@ def test_fail_add_player_to_event_already_in_it(self, client, saved_game_event,
...
+        mock_response.status = 200
+        mock_response.read.return_value = bytes('{"weather": [{"icon": "04d"}],"main": {"temp": 13.19}}', 'utf-8')
+        mock_urlopen.return_value = mock_response
+
+        client.force_login(player.user)
+
+        for index in range(len(saved_game_events_list)):
+            output = get_weather_from_api([saved_game_events_list[index]], mock_urlopen)
+            assert output[saved_game_events_list[index].id] == {'icon': "http://openweathermap.org/img/wn/04d@2x.png",
+                                                                'temp': 13.19}
```
yftacherzog, line 348: This test is not testing 2 entries. It tests calling the function twice with a different entry.

I'd suggest that instead of having a fixture for the list of events, you add another fixture for the second event and construct the list here. e.g. `saved_game_event, another_saved_game_event`
```suggestion
        output = get_weather_from_api([saved_game_event, another_saved_game_event], mock_urlopen)
        expected_weather = {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
        assert output == {
                saved_game_event.id: expected_weather,
                another_saved_game_event: expected_weather,
        }
        for index in range(len(saved_game_events_list)):
            output = get_weather_from_api([saved_game_events_list[index]], mock_urlopen)
            assert output[saved_game_events_list[index].id] == {'icon': "http://openweathermap.org/img/wn/04d@2x.png",
                                                                'temp': 13.19}
```

game_event/tests.py
```diff
@@ -265,3 +296,53 @@ def test_fail_add_player_to_event_already_in_it(self, client, saved_game_event,
...
+        mock_response = mock.Mock()
+        mock_response.status = 185
+        mock_urlopen.return_value = mock_response
+
+        client.force_login(player.user)
+
+        output = get_weather_from_api([saved_game_event], mock_urlopen)
+        assert output[saved_game_event.id] is None
+
+    def test_empty_game_events_list_valid_empty_dictionary_response(self, client, player, mock_urlopen):
```
yftacherzog, line 332:
```suggestion
    def test_empty_game_events_list_valid_empty_dictionary_response(self, client, player):
```

game_event/views.py
```diff
@@ -46,7 +49,7 @@ def process_game_event_form(request):
 
 def game_events(request):
     game_events = GameEvent.objects.all()
-
+    game_events_with_weather = []
```
yftacherzog, line 52: Variables should be declared as closest to where they are used. In this case, it should be just before the loop.

game_event/views.py
```diff
@@ -63,7 +66,20 @@ def game_events(request):
...
+        game_events_with_weather.append({
+            'id': game.id,
+            'time': game.time,
+            'level_of_game': game.level_of_game,
+            'min_number_of_players': game.min_number_of_players,
+            'max_number_of_players': game.max_number_of_players,
+            'court': game.court,
+            'ball_game': game.ball_game,
+            'weather': weather[game.id]
+        })
```
yftacherzog, line 80: I have some reservations about this implementation:
- We're changing the type of what is being passed along. It used to be whatever `GameEvent.objects.all()` is returning, and now it's a list - I'd prefer the type to be untouched.
- The elements themselves used to be game events and now they are dictionaries - I'd prefer them to stay game events.
- We assume we know all fields a game event should have, but this might not be true, and suppose we add another in the future, it will be hard to figure out that it's not showing because this function is discarding it, because it was unknown when it was written.

All in all, I think we should have the function that gets the weather to replace the events there with events with weather (maybe you'd need to add weather to the model for that - not sure):
```python
# rename to something like get_events_with_weather
def get_weather_from_api(game_events, weather_getter): 
    for game in game_events:
        ...
        ...
        # have the logic here that gets the weather or None and stores it into a variable called weather
        game.weather = weather
    return game_events
```

game_event/views.py
```diff
@@ -150,3 +166,33 @@ def process_answer_game_event(request, id):
...
+            response = weather_getter(full_url)
+        except Exception:
+            weather[game.id] = None
+            continue
+
+        if response.status == 200:
+            data = response.read().decode('utf-8')
+            weather_data = json.loads(data)
+            currIcon = weather_data['weather'][0]['icon']
+            iconUrl = "http://openweathermap.org/img/wn/" + currIcon + "@2x.png"
```
yftacherzog, line 192:
```suggestion
            icon_url = "http://openweathermap.org/img/wn/" + currIcon + "@2x.png"
```

yftacherzog (changes requested):
I left some comments. Let me know if something is unclear.

game_event/tests.py
IdanZimi: @yftacherzog Can you please elaborate on this? 

> (imagine a regression in which the function injects a trailing entry

I'm not sure I understand the scenario.

yftacherzog: Your function generates an output which is a dictionary. Currently, you're assuming that it contains only the entry you expect to be generated. It might be that it's generating some other entries as well. For that reason, it's better to assert on the entire output rather than specific portions of it - when possible.

Suppose that you expect the output to be:
```python
{
  "x": 5
}
```
You test it with:
```python
assert output["x"] == 5
```

Tomorrow I make a change that introduces a regression that makes it return
```python
{
  "x": 5,
  "super-secret-stuff": "super-secret",
}
```
Your test still passes.

I suggest you change the test so it asserts the complete output:
```python
assert output == {"x": 5}
```

IdanZimi: Got it. thank you.
Will be fixed @yftacherzog

game_event/tests.py
OhadSeg: @yftacherzog Went with your advice. Fixed.

game_event/tests.py
OhadSeg: Changed

game_event/tests.py
OhadSeg: @yftacherzog Fixed.

game_event/views.py
OhadSeg: @yftacherzog Fixed

game_event/tests.py
OhadSeg: @yftacherzog Removed all.

game_event/tests.py
OhadSeg: @yftacherzog @IdanZimi Fixed

game_event/views.py
OhadSeg: @yftacherzog Fixed.

game_event/views.py
OhadSeg: @yftacherzog Fixed

game_event/views.py
OhadSeg: @yftacherzog  We @IdanZimi  changed some stuff in accordance with your comment, but we did skip changing the model due to time schedule.

game_event/tests.py
OhadSeg: fixed

game_event/views.py
```diff
@@ -150,3 +154,38 @@ def process_answer_game_event(request, id):
...
+            'q': game.court.city,
+            'units': 'metric',
+            'appid': os.getenv("APPID")
+        }
+
+        full_url = url + '?' + urllib.parse.urlencode(params)
+        try:
+            response = weather_getter(full_url)
+        except Exception:
+            weather.append(None)
```
yftacherzog, line 173:
```suggestion
            event.weather = None
```

game_event/views.py
```diff
@@ -150,3 +154,38 @@ def process_answer_game_event(request, id):
...
+            continue
+
+        if response.status == 200:
+            data = response.read().decode('utf-8')
+            weather_data = json.loads(data)
+            curr_icon = weather_data['weather'][0]['icon']
+            icon_url = "http://openweathermap.org/img/wn/" + curr_icon + "@2x.png"
+            curr_temp = weather_data['main']['temp']
+            weather.append({'icon': icon_url, 'temp': curr_temp})
```
yftacherzog, line 182:
```suggestion
           event.weather = {'icon': icon_url, 'temp': curr_temp}
```

game_event/views.py
```diff
@@ -150,3 +154,38 @@ def process_answer_game_event(request, id):
...
+        else:
+            weather.append(None)
+
+    return weather
+
+
+def add_weather_to_game_event(weather_list, game_events):
+    for index, game_event in enumerate(game_events):
+        game_event.weather = weather_list[index]
```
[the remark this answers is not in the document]
yftacherzog: Also, if we add this function, we need to add tests for it.

game_event/tests.py
```diff
@@ -252,3 +273,47 @@ def test_fail_add_player_to_event_already_in_it(self, client, game_event, player
...
+        output = get_weather_from_api([game_event], mock_urlopen)
+        expected_weather = {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
+        assert output == [expected_weather]
+
+    def test_exception_thrown_from_api_call(self, game_event, player, mock_urlopen):
+
+        mock_urlopen.side_effect = Exception()
+
+        output = get_weather_from_api([game_event], mock_urlopen)
+        assert output[0] is None
```
yftacherzog, line 294:
```suggestion
        assert output == [None]
```

game_event/tests.py
```diff
@@ -252,3 +273,47 @@ def test_fail_add_player_to_event_already_in_it(self, client, game_event, player
...
+
+    def test_one_invalid_response_status_code(self, game_event, player, mock_urlopen):
+        mock_response = mock.Mock()
+        mock_response.status = 185
+        mock_urlopen.return_value = mock_response
+
+        output = get_weather_from_api([game_event], mock_urlopen)
+        assert output[0] is None
```
yftacherzog, line 302:
```suggestion
        assert output == [None]
```

yftacherzog (changes requested):
I think we should just modify the game event once we fetch the weather instead of finding a way to store it and locate the place to store it afterwards.

I added a couple of suggestions that illustrate that.

game_event/views.py
yftacherzog: If it allows the assignment of the weather attribute without changing the model then I completely ok with not changing the model.

yftacherzog: See my new comments on assigning the weather to the game event object directly rather than storing it on a temp list.

game_event/views.py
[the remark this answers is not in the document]
IdanZimi: Do you think we should remove this function and think of another way to implement? Or we can proceed with this implementation even though its fragile?  @yftacherzog

game_event/views.py
```diff
@@ -63,6 +65,8 @@ def game_events(request):
         current_date = date.today()
         game_events = game_events.filter(time__gte=current_date)
 
+    weather = get_weather_from_api(game_events, urllib.request.urlopen)
+    add_weather_to_game_event(weather, game_events)
```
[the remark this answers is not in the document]
IdanZimi: Great idea. We'll be fixed.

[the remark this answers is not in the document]
IdanZimi: Great idea. Will be fixed.

game_event/views.py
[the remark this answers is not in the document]
yftacherzog: The code suggestions I posted are for that.

game_event/views.py
```diff
@@ -150,3 +154,38 @@ def process_answer_game_event(request, id):
...
+
+
+def get_weather_from_api(game_events, weather_getter):
+    weather = []
+    for game in game_events:
+        url = 'https://api.openweathermap.org/data/2.5/weather'
+        params = {
+            'q': game.court.city,
+            'units': 'metric',
+            'appid': os.getenv("APPID")
```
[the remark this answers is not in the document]
yftacherzog: @SchSeba in case it's empty, it will just fail fetching the weather (there's a test case for that). This is also what allows running the app without providing this env var (i.e. it will work as usual but will not populate weather details).

I don't think there's an added value with checking the env var.

game_event/views.py
OhadSeg: @yftacherzog Fixed.

game_event/tests.py
```diff
@@ -252,3 +273,48 @@ def test_fail_add_player_to_event_already_in_it(self, client, game_event, player
...
+@pytest.mark.django_db
+class TestGetWeatherFromApi:
+    def test_one_valid_response(self, game_event, player, mock_urlopen):
+        mock_response = mock.Mock()
+        mock_response.status = 200
+        mock_response.read.return_value = bytes('{"weather": [{"icon": "04d"}],"main": {"temp": 13.19}}', 'utf-8')
+        mock_urlopen.return_value = mock_response
+        output = get_weather_from_api([game_event], mock_urlopen)
+        game_event.weather = {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
+        assert output == [game_event]
```
yftacherzog, line 287: I think you might be actually comparing the same thing here. The function is mutating the game event object it takes as input and returns it. Then you're using another reference of the same object and compare between the two.

To confirm, you can drop the line of assigning to `game_event.weather` and see if the assertion still passes. If it does, you need to compare against some other object.

game_event/tests.py
```diff
@@ -252,3 +273,48 @@ def test_fail_add_player_to_event_already_in_it(self, client, game_event, player
...
+
+    def test_exception_thrown_from_api_call(self, game_event, player, mock_urlopen):
+
+        mock_urlopen.side_effect = Exception()
+
+        output = get_weather_from_api([game_event], mock_urlopen)
+        assert output[0].weather is None
```
yftacherzog, line 294: My comment remains regarding asserting the entire collection rather just the one element.

game_event/tests.py
```diff
@@ -252,3 +273,48 @@ def test_fail_add_player_to_event_already_in_it(self, client, game_event, player
...
+        mock_response.read.return_value = bytes('{"weather": [{"icon": "04d"}],"main": {"temp": 13.19}}', 'utf-8')
+        mock_urlopen.return_value = mock_response
+
+        output = get_weather_from_api([game_event, another_game_event], mock_urlopen)
+        game_event.weather = {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
+        another_game_event.weather = {'icon': "http://openweathermap.org/img/wn/04d@2x.png", 'temp': 13.19}
+        assert output == [
+          game_event,
+          another_game_event
+        ]
```
yftacherzog, line 320: Same comment here about the two sides of the equation being references to the same object.

game_event/views.py
```diff
@@ -150,3 +154,32 @@ def process_answer_game_event(request, id):
...
+            data = response.read().decode('utf-8')
+            weather_data = json.loads(data)
+            curr_icon = weather_data['weather'][0]['icon']
+            icon_url = "http://openweathermap.org/img/wn/" + curr_icon + "@2x.png"
+            curr_temp = weather_data['main']['temp']
+            game.weather = {'icon': icon_url, 'temp': curr_temp}
+        else:
+            game.weather = None
+
+    return game_events
```
Yuval-Vino, line 185: No need to return game_events since the method modifies the input list directly.

game_event/views.py
```diff
@@ -150,3 +154,32 @@ def process_answer_game_event(request, id):
     content['event_exists'] = True
 
     return render(request, 'game_event/result.html', content)
+
+
+def get_weather_from_api(game_events, weather_getter):
+    for game in game_events:
+        url = 'https://api.openweathermap.org/data/2.5/weather'
```
Yuval-Vino, line 161: no reason to create the same URL each iteration.
```suggestion
    url = 'https://api.openweathermap.org/data/2.5/weather'
    for game in game_events:
        
```


    
also true for the `params `variable
You can define the dictionary with the common values, `units `and `appid`, before the for loop since they are the same for all game_events. You can change the variable `params['q']` inside the loop based on the specific game event's `court.city`."

This change improves the readability of the function

game_event/tests.py
OhadSeg: Hey @yftacherzog you were right we we do compare the same thing.
What other object we can compare to, the problem is that the function we are testing returns a list of game events, so we need to compare a list with another list, but the problem is that we cannot create another game_event object with the same id. wdyt how can we fix it ?

game_event/views.py
yftacherzog: It's better to return it. Functions that rely on side effects on the inputs are a mess.

game_event/tests.py
yftacherzog: In that case, you can fall back to asserting on the weather directly rather than the complete list. We tried testing more thoroughly. Didn't work.

game_event/views.py
Yuval-Vino: > It's better to return it. Functions that rely on side effects on the inputs are a mess.

I hadn't thought about this, but doesn't it still rely on side effects since we're updating the input variable `game_event`?
if I understand correctly function rely on side effect when it changes something outside of its function definition so to make this function pure we will need to create `new_game_events `list which copies the `game_events `and then make the changes on `new_game_events`, or am i missing something @yftacherzog ?

yftacherzog: It's cleanest if you have an  immutable object and instead of mutating it, you return a new object.

The current approach is not ideal, but I find the code easier to handle if you return the things you mutated rather than change them only as a side effect.

game_event/tests.py
IdanZimi: We get back from the `get_weather_from_api` a list of game_events. from what I understand in that case the weather field will be `None `. Is `output == [None]` assertion checking that? 
`output[0].weather`  is the way to access the weather no? @yftacherzog

game_event/views.py
IdanZimi: fixed, no more use of append.

game_event/views.py
IdanZimi: We took out the constant `params ` - `units `, `url  `and `appid `out of the for loop.

game_event/tests.py
IdanZimi: _"In that case, you can fall back to asserting on the weather directly rather than the complete list. We tried testing more thoroughly. Didn't work"_

After that. Is the assertion meets you expectations? @yftacherzog

game_event/tests.py
yftacherzog: The comment was written when we were still returning just the weathers.
We know that we're limited in our ability to compare against the entire object, but at least I'd like us to get some notion that the list is containing the stuff that we expect and only that.

With that in mind, you can do something like:
```python
weathers = [e.weather for e in output]
assert weathers = [None]
```
It's not perfect, but it's verifying that we're getting the expected number of elements.

You can do something similar also when you have more than one instance.

game_event/views.py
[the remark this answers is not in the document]
yftacherzog: @OhadSeg we should let @SchSeba resolve his own comment.

game_event/tests.py
yftacherzog: Yes. That should do.

yftacherzog (approved):

game_event/views.py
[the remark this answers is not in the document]
OhadSeg: ok my bad @SchSeba @yftacherzog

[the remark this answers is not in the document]
SchSeba: works for me

SchSeba (approved):
LGTM

Yuval-Vino (approved):
Great job on addressing all the comments and going through multiple review cycles(surely it wasn't easy)!
LGTM!

PR #85, opened 2023-05-22
Add court and court_ball_game to conftest

amitKaplansky:
### Desired Outcome

one fixture of court and court_ball_game, reducing the creation of those entities in test files.
Will make the code more readable.

### Implemented Changes

-  court and court_ball_game fixtures added to conftest.
-  Adjustments were made in the relevant test files for change to work: remove court and court_ball_game from the test files, changed the name from saved_ court => court, court_ball_game_instance / saved_court_ball_game => court_ball_game.
- Removed saved_player from game_event tests, player already in contest no need for the same fixture in that file.

### Connected Issue/Story

Resolves #80 

### Dependencies

- [x] This PR does not depend on other PR's 

### Definition of Done
one court and court_ all_game fixture, that all the other test files will use if needed

#### Test coverage

- [x] The changes in this PR do not require tests

#### Documentation

- [x] Most of the test files required adjustments.

PR #96, opened 2023-05-27
Add chat to game-event page

### LINKED ISSUE #95: Add chat to game-event page
### **Expected Behavior**
On game-event page, should be a chat to the specific event.

### **Current Behavior**
not implemented

MaayanMashhadi:
### Desired Outcome

Players will be able to chat with each other in the specific event. Only the players
who joined the event, will be able to enter the chat.

### Implemented Changes

1. modify game-event html for adding chat html
2. modify game-event view to present the messages
3. adding new tests 
4. adding new view function for chat
5. adding new url for the view function

### Connected Issue/Story

Resolves #95 

### Dependencies

#### Test coverage

  changes

#### Documentation

### DIFF
--- a/game_event/templates/game_event/game-event.html
+++ b/game_event/templates/game_event/game-event.html
@@ -4,70 +4,140 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <link href="/static/css/chat.css" rel="stylesheet" />
+      
     <title>Game Event</title>
     <link rel="stylesheet" type="text/css" href="{% static 'css/game-event.css' %}">
-    <script>
-        setTimeout(function() {
-            location.reload();
-        }, 5000);
-    </script>
+    
   </head>
   <body>
-    {% if time %}
-        <h1>Event Details</h1>
-        <p>Time: {{ time }}</p>
-        <p>Level: {{ level_of_game }}</p>
-        <p>Min Players: {{ min_number_of_players }}</p>
-        <p>Max Players: {{ max_number_of_players }}</p>
-        <p>Court: {{ court }} {{ neighborhood }}</p>
-        <p>Game: {{ ball_game }}</p>
-        <h2>Players:</h2>
-        <table class="nice-table">
-            <thead>
-                <tr>
-                <th> </th>
-                <th>Name</th>
-                <th>Brings Ball</th>
-            </tr>
-            </thead>
-                <tbody>
-                    {% for player in event_players %}
-                        <tr>
-                            <td>
-                                <img src="{% static 'images/favicon.ico' %}" alt="Ball Image" class="ball-image">
-                            </td>
-                            <td>{{ player.first_name }} {{ player.last_name }}</td>
-                            <td>
-                                {% if player.brings_ball == True %}
-                                    Yes
+        {% if time %}
+            <h1>Event Details</h1>
+            <p>Time: {{ time }}</p>
+            <p>Level: {{ level_of_game }}</p>
+            <p>Min Players: {{ min_number_of_players }}</p>
+            <p>Max Players: {{ max_number_of_players }}</p>
+            <p>Court: {{ court }} {{ neighborhood }}</p>
+            <p>Game: {{ ball_game }}</p>
+            <h2>Players:</h2>
+            <table class="nice-table">
+                <thead>
+                    <tr>
+                    <th> </th>
+                    <th>Name</th>
+                    <th>Brings Ball</th>
+                </tr>
+                </thead>
+                    <tbody>
+                        {% for player in event_players %}
+                            <tr>
+                                <td>
+                                    <img src="{% static 'images/favicon.ico' %}" alt="Ball Image" class="ball-image">
+                                </td>
+                                <td>{{ player.first_name }} {{ player.last_name }}</td>
+                                <td>
+                                    {% if player.brings_ball == True %}
+                                        Yes
+                                    {% else %}
+                                        No
+                                    {% endif %}
+                                </td>
+                            </tr>
+                        {% endfor %}
+                    </tbody>
+                </table>
+                <div>
+                    {% if in_event %}
+                        <a href="/game-events/remove-from-event/{{ id }}/"
+                        class="round-button"
+                        onclick="window.open(this.href, 'Remove Event', 'width=400,height=250,left='+((screen.width-400)/2)+',top='+((screen.height-515)/2)); return false;">
+                        Remove Event
+                        </a>
+                            <div class="top-right">
+                                <h1>Chat With Players!</h1>
+                            <body>
+                                    
+                                    {% if all_messages %}
+                                    <div id="savedText">
+                                            <div class="message-box">
+                                              {% for message in all_messages %}
+
+                                              {% if message.user_id.user.username == request.user.username %}
+                                                <div class="message-container">
+                                                  <span class="message-time">{{ message.time_sent|date:"F d, H:i" }}</span> 
+                                                  <span class="name">{{ message.user_id.user.username }}</span><br>
+                                                            <p><span class="message">{{ message.text }}</span></p>
+                                                </div>
+
+                                                {% else %}
+                                                <div class="message-other-user-container">
+                                                  <span class="message-other-time">{{ message.time_sent|date:"F d, H:i" }}</span> 
+                                                  <span class="name-other">{{ message.user_id.user.username }}</span><br>
+                                                            <p><span class="message-other">{{ message.text }}</span></p>
+                                                </div>
+                                                {% endif %}
+                                              {% endfor %}
+                             
+                                              
+                                            </div>
+                                          </div>
+                                        
                                 {% else %}
-                                    No
+                                    <div id="savedText" style="display: none;">
+                <textarea id="savedTextbox" rows="12" cols="50" readonly></textarea>
+         </div>
                                 {% endif %}
-                            </td>
-                        </tr>
-                    {% endfor %}
-                </tbody>
-            </table>
+                                
+                                <div class="input">
+                                    <form method="POST" action="{% url 'save_text_message' id=id %}">
+                                        {% csrf_token %}
+                                    <input type="text" id="mytextbox" name="mytextbox" placeholder="Type your message here!">
+                                    <button onclick="saveText(); showSaveBox();">Send</button>
+                                    </div>
+                                    </form>
+                                <script>
+                                    function saveText() {
+                                        var textbox = document.getElementById('mytextbox');
+                                        var savedTextbox = document.getElementById('savedTextbox');
+                                        
+                                    }
+                                    function showSaveBox() {
+                                        var savedTextDiv = document.getElementById('savedText');
+                                        savedTextDiv.style.display = 'block';
+                                     }
+                                     
+                                </script>
+                                {% if messages %}
+                                <div class="row my-4">
+                                  <div class="col">
+                                    <div class="alert alert-danger" role="alert">
+                                      <ul class="messages">
+                                        {% for message in messages %}
+                                          <li{% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
+                                        {% endfor %}
+                                      </ul>
+                                    </div>
+                                  </div>
+                                </div>
+                              {% endif %}
+
+                            </body>
+							
+                            
+                            </div>
+                    {% else %}
+                        <a href="/game-events/join-event/{{ id }}/"
+                        class="round-button"
+                        onclick="window.open(this.href, 'Join Event', 'width=400,height=250,left='+((screen.width-400)/2)+',top='+((screen.height-515)/2)); return false;">
+                        Join Event
+                        </a>
+                    {% endif %}
+                </div>
+        {% else %} 
             <div>
-                {% if in_event %}
-                    <a href="/game-events/remove-from-event/{{ id }}/"
-                    class="round-button"
-                    onclick="window.open(this.href, 'Remove Event', 'width=400,height=250,left='+((screen.width-400)/2)+',top='+((screen.height-515)/2)); return false;">
-                    Remove Event
-                    </a>
-                {% else %}
-                    <a href="/game-events/join-event/{{ id }}/"
-                    class="round-button"
-                    onclick="window.open(this.href, 'Join Event', 'width=400,height=250,left='+((screen.width-400)/2)+',top='+((screen.height-515)/2)); return false;">
-                    Join Event
-                    </a>
-                {% endif %}
+                <h1> Event Does not exist. </h1>
             </div>
-       {% else %} 
-        <div>
-            <h1> Event Does not exist. </h1>
-        </div>
-       {% endif %}
-  </body>
+        {% endif %}
+ </body>
 </html>
 {%endblock%}
--- a/message/models.py
+++ b/message/models.py
@@ -1,7 +1,9 @@
 from django.db import models
 from player.models import Player
 from game_event.models import GameEvent
+from game_event_player.models import GameEventPlayer
 from django.core.validators import MinLengthValidator
+from django.core.exceptions import ValidationError
 
 
 class Message(models.Model):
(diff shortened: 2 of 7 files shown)


MaayanMashhadi:
New tests will be uploaded in the next days.

Danielsio:
Keep up the good work,  is there screenshots that we can see in the meantime ?

MaayanMashhadi:
This is the chat:
chat

this is the page:
the page

MaayanMashhadi:
> Keep up the good work, is there screenshots that we can see in the meantime ?

did it

ZvikaNaorCohen:
Great job, the chat looks awesome! I really like it.

message/tests.py
```diff
@@ -14,41 +18,63 @@
...
+@pytest.fixture
+def user2():
+    return get_user_model().objects.create_user(
+        username='testuser1',
+        password='testpass1'
+    )
+
+
+@pytest.fixture
+def saved_player(user):
```
ZvikaNaorCohen, line 38: I think that on PR #85 we moved all the fixtures to conftest.
Can we use the conftest file instead of making a new fixture here?
Maybe even adding there `saved_player2` that is written below.

game_event/templates/game_event/game-event.html
```diff
@@ -4,70 +4,291 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <style>
+      @import url('https://fonts.googleapis.com/css?family=Red+Hat+Display:400,500,900&display=swap');
```
SchSeba, line 8: please put  this into a css file

SchSeba (commented):
looks good just a small comment

btw if you have time to implement and not part of this PR
you can improve the chat a bit having a 5 sec loop to request the messages (no need to implement a full web socket chat) that will ask the server for messages and reload just the chat box with new messages so you don't need to refresh the page everytime

message/tests.py
MaayanMashhadi: > I think that on merged PR #85 we moved all the fixtures to conftest. Can we use the conftest file instead of making a new fixture here? Maybe even adding there `saved_player2` that is written below.

You're right, I did it

game_event/templates/game_event/game-event.html
MaayanMashhadi: > please put this into a css file

Done

amitCohen2:
Great job!
The tests and all looks great for me! Didn't find any comments to write about it.

message/views.py
```diff
@@ -1,3 +1,33 @@
...
-# Create your views here.
+
+def save_text_message(request, id):
+    try:
+        event = GameEvent.objects.get(pk=id)
+    except GameEvent.DoesNotExist:
+        return render(request, 'game_event/game-event.html', {})
+    try:
+        player = Player.objects.get(user=request.user)
+    except Player.DoesNotExist:
```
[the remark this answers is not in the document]
Danielsio: i would still be able to alter the id value at the url of the browser if i am not wrong.
so the login decorator is a good addition but not an alternative

Danielsio (approved):

message/views.py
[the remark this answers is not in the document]
Yuval-Vino: @Danielsio If I understand correctly, the `id `referred to here is the event ID, not the user ID (which is obtained from the URL). However, you raise a valid concern that a user could potentially modify the event ID in the URL and send a message to an event they are not actually a part of. Perhaps it would be wise to implement a verification step before sending the message to ensure that the player is indeed a member of the event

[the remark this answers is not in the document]
MaayanMashhadi: > @Danielsio If I understand correctly, the `id `referred to here is the event ID, not the user ID (which is obtained from the URL). However, you raise a valid concern that a user could potentially modify the event ID in the URL and send a message to an event they are not actually a part of. Perhaps it would be wise to implement a verification step before sending the message to ensure that the player is indeed a member of the event

Hi, just see your comment after I push the changes. So do you think I need to do some verification about it?

message/tests.py
```diff
@@ -80,27 +70,97 @@ def test_delete_user_deletes_message(self, saved_message):
...
+    def test_success_send_message(self, client, saved_game_event, player):
+        client.force_login(player.user)
+        game_event_id = saved_game_event.id
+        count_before = Message.objects.count()
+        url = reverse('save_text_message', args=[game_event_id])
+        response = client.post(url, {'mytextbox': 'Test message'})
+
+        assert response.status_code == 302
+        assert Message.objects.count() == count_before + 1
+        message = Message.objects.first()
```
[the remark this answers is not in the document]
MaayanMashhadi: Done

message/tests.py
```diff
@@ -14,41 +18,27 @@
 TEST_BALL_GAME = 'Basketball'
 
 
+@pytest.fixture
+def saved_game_event(court):
```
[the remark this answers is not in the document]
MaayanMashhadi: I used the fixture from conftest so I remove this fixture from this file.

message/tests.py
```diff
@@ -14,41 +18,27 @@
...
+    return GameEvent.objects.create(id=TEST_ID, time=TEST_TIME, level_of_game=TEST_LEVEL,
+                                    min_number_of_players=TEST_MIN, max_number_of_players=TEST_MAX,
+                                    court=court, ball_game=TEST_BALL_GAME)
+
+
+@pytest.fixture
+def saved_message(player, saved_game_event):
```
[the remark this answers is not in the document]
MaayanMashhadi: > consider renaming it to `message`

Done

message/tests.py
```diff
@@ -80,27 +70,97 @@ def test_delete_user_deletes_message(self, saved_message):
...
+        url = reverse('save_text_message', args=[game_event_id])
+        response = client.post(url, {'mytextbox': ''})
+
+        assert response.status_code == 302
+        assert Message.objects.count() == count_before
+
+    def test_failed_send_message_invalid_player(self, client, saved_game_event):
+        user_without_player = get_user_model().objects.create_user(username='user_without_player',
+                                                                   password='user_without_player1')
+        client.force_login(user_without_player)
```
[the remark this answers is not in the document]
MaayanMashhadi: > can we have a user without a player?

You're right, I remove this test. Maybe I think to add some verification that will ensure only players on this event have the permission to send messages on the chat, and I will upload new test about it.

message/views.py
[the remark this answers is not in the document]
MaayanMashhadi: I will add this verification, you're right.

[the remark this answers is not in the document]
Yuval-Vino: By using `@login_required`, we ensure that only logged-in players can access the app. The login and register parts already have tests, so we can rely on them.

But we need to verify the player's participation in the event.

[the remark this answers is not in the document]
MaayanMashhadi: > By using `@login_required`, we ensure that only logged-in players can access the app. The login and register parts already have tests, so we can rely on them.
> 
> But we need to verify the player's participation in the event.

I added verification and modify the test with invalid player

message/models.py
```diff
@@ -1,11 +1,36 @@
...
+                          game_event_id=game_event,
+                          text=text)
+
+        message.save()
+        return message
+
+    @staticmethod
+    def validate_message(player, game_event, text):
+        game_event = GameEvent.objects.get(pk=game_event.pk)
+        game_event_player = GameEventPlayer.objects.filter(player=player, game_event=game_event).exists()
```
yftacherzog, line 28: The variable name should imply its type. Currently, the reader would expect it to contain a player. Can rename to something like this:
```suggestion
        player_participates = GameEventPlayer.objects.filter(player=player, game_event=game_event).exists()
```

message/models.py
```diff
@@ -1,11 +1,36 @@
...
+        errors = []
+        if game_event_player is False:
+            errors.append("Just players from this event can send messages on chat")
+        if text == '':
```
yftacherzog, line 32: Also consider if you'd like to catch other types of empty messages - something like this:
```suggestion
        if text.strip()':
```

message/tests.py
```diff
@@ -14,93 +15,126 @@
...
-
-    @pytest.fixture
-    def saved_message(self, saved_player, saved_game_event):
-        message = Message.objects.create(
-            user_id=saved_player,
-            game_event_id=saved_game_event,
-            time_sent=TEST_TIME,
-            text="test message"
-        )
-        return message
```
yftacherzog, line 50: Are those fixture changes related to this PR?

yftacherzog (changes requested):
As far as I understand it, this PR includes quite a lot of refactoring that are not directly related to the functionality being introduced. Unless the refactoring is tightly coupled with the code being introduced, it's better to split them into different PRs. It makes it less likely that you'd have uncaught regressions and it also makes it easier to review which makes it less likely to introduce new bugs.

message/tests.py
MaayanMashhadi: > Are those fixture changes related to this PR?

Those fixtures are inside `class TestMessageModel`. If I not put those fixtures outside of this class,
my tests will not recognize those fixtures and I will duplicates those fixtures. But according to #102 it isn't necessary right now.
I will wait to merge #102 PR and fix it. About the other changes - you're right, I will open an issue about it and open ne PR for it.

MaayanMashhadi: > Are those fixture changes related to this PR?

About the position of those fixture I think yes, because those fixtures are inside the class `TestMessageModel`, and my new tests aren't recognize those fixtures because of it. About the renaming, you're right and I open an issue about it #106

MaayanMashhadi:
> As far as I understand it, this PR includes quite a lot of refactoring that are not directly related to the functionality being introduced. Unless the refactoring is tightly coupled with the code being introduced, it's better to split them into different PRs. It makes it less likely that you'd have uncaught regressions and it also makes it easier to review which makes it less likely to introduce new bugs.

I would like to know if you talked about just the changes of the fixtures I made or other things you think it needs to be in another PR? I think the other changes related to the chat features.

yftacherzog:
> 
> 
> 
> > As far as I understand it, this PR includes quite a lot of refactoring that are not directly related to the functionality being introduced. Unless the refactoring is tightly coupled with the code being introduced, it's better to split them into different PRs. It makes it less likely that you'd have uncaught regressions and it also makes it easier to review which makes it less likely to introduce new bugs.
> 
> I would like to know if you talked about just the changes of the fixtures I made or other things you think it needs to be in another PR? I think the other changes related to the chat features.

I don't recall anymore. I'll let you be the judge.

static/css/chat.css
```diff
@@ -0,0 +1,164 @@
...
+  font-weight: 400;
+  letter-spacing: 0.025em;
+  width: 4.5rem;
+  color:#333
+}
+
+button:hover {
+  background-color: #f3f3f3;
+  cursor: pointer;
+}
```
yftacherzog, line 164: add newline
```suggestion
}

```

message/models.py
```diff
@@ -1,11 +1,36 @@
...
+        Message.validate_message(player, game_event, text)
+        message = Message(user_id=player,
+                          game_event_id=game_event,
+                          text=text)
+
+        message.save()
+        return message
+
+    @staticmethod
+    def validate_message(player, game_event, text):
```
yftacherzog, line 26: We should have unit tests that tests this method directly. We give different combinations of inputs and assert on the output (None) or the content of the error message.

yftacherzog (changes requested):
Whenever we have an opportunity to test a unit (usually a function) on its own, we should prefer that over testing it inside something else.

We should also have that something-else tested, but then we should focus on its functionality rather than the functionality provided by the unit it utilize.

message/models.py
MaayanMashhadi: > We should have unit tests that tests this method directly. We give different combinations of inputs and assert on the output (None) or the content of the error message.

I added 3 tests:
 1. checks the validation is success when the input are correct
 2. checks the validation will raise 'ValidationError' when player that isn't on the event sends message,
 and checks the error message is correct
3. checks the validation will raise 'ValidationError' when player sends empty message/message that contains only space character, and checks the error message is correct

message/models.py
```diff
@@ -1,11 +1,36 @@
...
 
 
 class Message(models.Model):
     user_id = models.ForeignKey(Player, on_delete=models.CASCADE)
     game_event_id = models.ForeignKey(GameEvent, on_delete=models.CASCADE)
     time_sent = models.DateTimeField(auto_now_add=True)
     text = models.CharField(max_length=255, validators=[MinLengthValidator(1)])
+
+    @staticmethod
+    def create(player, game_event, text):
```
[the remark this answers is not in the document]
MaayanMashhadi: > We should have a unit test that tests this directly. No need for all the different failure options. Just one success and one failure.

I added 2 tests for success scenario and failure scenario (which player that isn't on the event try to sends message)

message/models.py
yftacherzog: Nice!

message/models.py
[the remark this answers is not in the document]
yftacherzog: cool

message/tests.py
```diff
@@ -81,3 +84,95 @@ def test_create_message_with_empty_text(self, player, game_event):
...
+        e = Message.validate_message(player, game_event, text)
+        assert e is None
+
+    @pytest.mark.parametrize(
+        "text, expected_errors",
+        [
+            ('test message', ["Just players from this event can send messages on chat"]),
+            ('', ["Message can't be empty"]),
+            ('           ', ["Message can't be empty"]),
+        ]
```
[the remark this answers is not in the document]
MaayanMashhadi: > As you already implemented here a way to verify multiple errors, I suggest you assert on them all. I think this should work. No?

Yes it works, I change a little bit the function to check if the two error messages are raises (players that doesn't in the event and sends empty message).

message/tests.py
```diff
@@ -81,3 +84,93 @@ def test_create_message_with_empty_text(self, player, game_event):
...
+        [
+            ('test message', ["Just players from this event can send messages on chat"]),
+            ('', ["Message can't be empty", "Just players from this event can send messages on chat"]),
+            ('           ', ["Message can't be empty", "Just players from this event can send messages on chat"]),
+        ]
+    )
+    def test_fail_validate_message(self, game_event, player, text, expected_errors):
+        with pytest.raises(ValidationError) as current_errors:
+            Message.validate_message(player, game_event, text)
+        errors = "\n".join(current_errors.value)
```
yftacherzog, line 152: What do we have in `current_errors.value`? Looking at what we're putting there, I would not expect there would be anything for which `join` is applicable.

MaayanMashhadi: > expect

the current_errors.value is this: ["Just players from this event can send messages on chat\nMessage can't be empty"]
on the validation function, I convert the list to single string, and the errors separated by '\n' (I used the join function there)
If on this test I don't use the join it did me this assertion error: 
`E   assert "Just players from this event can send messages on chat\nMessage can't be empty" in ["Message can't be empty", 'Just players from this event can send messages on chat']

message/tests.py:155: AssertionError`

MaayanMashhadi: > What do we have in `current_errors.value`? Looking at what we're putting there, I would not expect there would be anything for which `join` is applicable.

the current_errors.value is this: ["Just players from this event can send messages on chat\nMessage can't be empty"]
on the validation function, I convert the list to single string, and the errors separated by '\n' (I used the join function there)
If on this test I don't use the join it did me this assertion error:
`E assert "Just players from this event can send messages on chat\nMessage can't be empty" in ["Message can't be empty", 'Just players from this event can send messages on chat']

message/tests.py:155: AssertionError`

yftacherzog: The outcome of running join on that is just getting the first element of that list - which is the only element.
```python
In [2]: a = ["Just players from this event can send messages on chat\nMessage can't be empty"]

In [3]: "\n".join(a)
Out[3]: "Just players from this event can send messages on chat\nMessage can't be empty"

In [4]: "\n".join(a) == a[0]
Out[4]: True
```
In other words, you can use `[current_errors.value]` instead of what you put into `errors`

message/tests.py
```diff
@@ -81,3 +84,93 @@ def test_create_message_with_empty_text(self, player, game_event):
...
+            ('', ["Message can't be empty", "Just players from this event can send messages on chat"]),
+            ('           ', ["Message can't be empty", "Just players from this event can send messages on chat"]),
+        ]
+    )
+    def test_fail_validate_message(self, game_event, player, text, expected_errors):
+        with pytest.raises(ValidationError) as current_errors:
+            Message.validate_message(player, game_event, text)
+        errors = current_errors.value.messages[0].split("\n")
+        for error in errors:
+            assert error in expected_errors
```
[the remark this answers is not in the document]
MaayanMashhadi: > suppose that `errors` comes back empty for some reason, you won't fail here and you couldn't know it. Try doing something like this:

You're right, I added this line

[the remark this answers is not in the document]
yftacherzog: Cool. You don't need both this and the for-loop.

yftacherzog (approved):
lgtm. Good job!

Danielsio (approved):

Yuval-Vino (approved):

PR #99, opened 2023-05-27
Add player-event linking and event deletion

amitKaplansky:
Implement deletion of event when the last player leaves

### Desired Outcome

- Game events should not be left empty. When the last player leaves an event, the event should be automatically deleted.
- After creating a game event, the player who created the event will be automatically added to it.
- Logging in will be required for the player to create a new event.

*Please describe the desired outcome for this PR.  Said another way, what was
the original request that resulted in these code changes?*

- Game events can currently created without any players being linked to them.
- When the last player leaves an event, the event is not automatically deleted.
- Currently, after creating a game event, the player who created the event is not automatically joined to it.

### Implemented Changes

- Modify GameEvent model, There is a conflict between the id field provided by the built-in Django framework and the id field defined in the mode, the id Field is now deleted from the model.
- Add a migration named "0003_alter_gameevent_id" for the "game_event" app after running the "makemigrations" command, This migration will alter the "id" field in the "game_event" model.
- Modify `remove_from_event` method.
- Modify `process_game_event_form`
- Added 2 methods to the GamEvent model.
- Modify tests of creation of game event, `client.force_login(player.user)`:  player most be logged in to create an event.

### Connected Issue/Story

Resolves #97 
Resolves #98 

### Dependencies

- [x] This PR does not depend on other PR's 

### Definition of Done
1. Modify `remove_from_event` method.
2. Modify `process_game_event_form`
3. Added two methods to the GamEvent model.
4. Modify tests of creation of game event, `client.force_login(player.user)`:  player most be logged in to create an event.
5. Implement a test to verify that the player has been successfully added to the event after its creation.
6.  Implement a test to ensure that the game event has been deleted correctly and that the corresponding page no longer exists.

#### Test coverage

- [x] This PR includes new unit and integration tests to go with the code
  changes

PR #107, opened 2023-06-02
Modify game event template, Implement logic on join event

### LINKED ISSUE #41: Add logic to Time in Game Event entity - Time
Time of an event can't be before the time it was created.
In addition,  the expected time of a game event most be a least 4 hours or more after for the time it was created.

### LINKED ISSUE #108: Implement logic on join event
Expected Behavior
Player Can Join Event Based on Availability and Time Gap

Current Behavior

The system does not enforce any restrictions based on overlapping events, and the player can join multiple events simultaneously.

### LINKED ISSUE #109: Modify Game event template
Expected Behavior

Game event template will not need the refresh script as before, we had conflicts with the chat and the refresh.

Current Behavior

After a player joins an event, a new window automatically opens to provide a confirmation message indicating whether the joining process was successful or not.
Additionally, the main page where the player initiated the joining process undergoes an automatic refresh after 5 seconds.
The purpose of the auto-refresh is to update the player's view and check if their joining status has changed, to check it join event button needs to be presented or remove event.

amitkaplansky:
### Desired Outcome

* Game event template will not need the refresh script as before, we had conflicts with the chat and the refresh.
* No result.html every change will we shown in the game_event.html.
* Add logic to join event.

### Implemented Changes

* Added messages error and success to game_event html.
* modified join_event, remove_from_event, process_answer_game_event in views.
* removed result html
* Added is_event_time_available in player model to check if player can join the event.
* Added style to game_event css for presenting error and success messages.
* Added two tests in `game_event/tests.py`
* Added two fixtures to conftest.

* removed id field from game_event model, added the new migration `0003_alter_gameevent_id.py`, it will be removed after #99 is merged. 

### Connected Issue/Story

Resolves #108
Resolves #109 
Resolves  #41 

### Dependencies

- [x] This PR depends on #99.

### Definition of Done
* No need for refresh on `game_event.html`.
* Show relevant messages in `game_event.html`.
* Player can join an event only if he is 'free'.

#### Test coverage

- [x] This PR includes two unit and integration tests to go with the code


amitKaplansky:
https://github.com/redhat-beyond/MeetBalls/assets/106595737/bf841a67-5e4a-492b-9958-4afefe5e6830

You can take a look at the changes (:

game_event/tests.py
Danielsio, line 1: can you explain the usage of the "follow' attribute in the calls ? 
why it's required?

Danielsio:
i think you can reference  #41  issue as well

game_event/models.py
```diff
@@ -51,3 +50,9 @@ def validate_game_event(court, ball_game, min_number_of_players, max_number_of_p
     @staticmethod
     def is_ball_game_playable_at_court(court, ball_game):
         return CourtBallGame.is_ball_game_playable(court, ball_game)
+
+    def is_event_full(self):
+        from game_event_player.models import GameEventPlayer
```
ZvikaNaorCohen, line 55: Can we move this import to the beginning of the file?

game_event/templates/game_event/join-event.html
```diff
@@ -3,6 +3,7 @@
 <html>
   <head>
     <link rel="stylesheet" type="text/css" href="{% static 'css/game-event.css' %}">
+    <link rel="stylesheet" type="text/css" href="{% static 'css/game-events.css' %}">
```
ZvikaNaorCohen, line 6: I can't see file `static/css/game-events.css` in this PR. Perhaps this line is redundant?

ZvikaNaorCohen:
@amitKaplansky and @MaayanMashhadi , the video looks awesome! I really like the new features you've added.
I left some small comments. Looks very good to me.

game_event/templates/game_event/game-event.html
```diff
@@ -2,72 +2,89 @@
...
+                    </td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+    <div class="button-container">
+        {% if in_event %}
+        <a href="/game-events/remove-from-event/{{ id }}" class="round-button remove-event-button">
+            Remove Event
```
Yuval-Vino, line 66: Remove event sounds like the event is being removed and not the player-event
```suggestion
            Leave Event
```

Yuval-Vino (commented):
I think it will be better to disable the 'Join Event' button when a player is unable to join the event? Instead of rendering the 'Are you bringing a ball?' message and only then receiving the 'Can't join' error.

game_event/templates/game_event/game-event.html
```diff
@@ -2,72 +2,89 @@
...
+    <div>
+        <h1> Event Does not exist. </h1>
+    </div>
+    {% endif %}
+
+</body>
+
 </html>
-{%endblock%}
+{%endblock%}
```
SchSeba, line 90: please add a new line here

game_event/views.py
```diff
@@ -108,45 +134,36 @@ def game_event(request, id):
...
-        content["reached_max"] = False
+    if GameEventPlayer.objects.filter(game_event=event, player=player):
+        messages.error(request, "Already in event!")
+        return redirect(f"/game-events/{id}", add_content(event, player))
+
+    is_event_full = event.is_event_full()
+    is_event_time_available = player.is_event_time_available(event.time)
+
+    error_messages = []
+    if is_event_full:
```
SchSeba, line 156: I think this validation should be on the even page not after we click on the join

player/models.py
```diff
@@ -54,3 +55,13 @@ def validate_and_save(self, birth_date, favorite_ball_game):
 
         else:
             self.save()
+
+    def is_event_time_available(self, desired_event_time):
+        from game_event_player.models import GameEventPlayer
```
SchSeba, line 60: please move the import no need todo import inside a function

game_event/models.py
amitkaplansky: No, It will cause a circular dependency

player/models.py
amitkaplansky: it will cause a circular dependency between the player model and the game_event_player

